PR #13, opened 2016-11-28
Reuse code between callback and promise code lines. Some other minor tweaks.

scriby:
I didn't take the time to test the changes manually, but it does still build and pass the unit tests.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -350,7 +350,7 @@ export class GraphRequest {
     // Given the built SuperAgentRequest, get an auth token from the authProvider, make the request and return a promise
     private routeResponseToPromise(requestBuilder:request.SuperAgentRequest) {
         return new Promise((resolve, reject) => {
-            routeResponseToCallback(requestBuilder, (err, body) => {
+            this.routeResponseToCallback(requestBuilder, (err, body) => {
                 if (err != null) {
                     reject(err);
                 } else {
--- a/src/index.ts
+++ b/src/index.ts
@@ -347,40 +347,28 @@ export class GraphRequest {
         );
     }
 
-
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -392,10 +380,10 @@ export class GraphRequest {
      */
     private sendRequestAndRouteResponse(requestBuilder:request.SuperAgentRequest, callback?:GraphRequestCallback):Promise<any>|void {
         // return a promise when Promises are supported and no callback was provided
-        if (typeof callback === "undefined" && typeof Promise !== "undefined") {
+        if (callback == null && typeof Promise !== "undefined") {
             return this.routeResponseToPromise(requestBuilder);
         } else {
-            this.routeResponseToCallback(requestBuilder, callback);
+            this.routeResponseToCallback(requestBuilder, callback || function(){});
```
scriby: Pass a dummy function in so we don't have to worry about whether it exists or not in routeResponseToCallback.

src/index.ts
```diff
@@ -392,10 +380,10 @@ export class GraphRequest {
      */
     private sendRequestAndRouteResponse(requestBuilder:request.SuperAgentRequest, callback?:GraphRequestCallback):Promise<any>|void {
         // return a promise when Promises are supported and no callback was provided
-        if (typeof callback === "undefined" && typeof Promise !== "undefined") {
+        if (callback == null && typeof Promise !== "undefined") {
```
scriby: Handle both null and undefined cases

dan-silver:
@scriby  This looks great!  Before I merge this architectural change in that deals with callbacks and promise routing, I'll need to add some more tests there. Right now, the tests are only around the chained query builder. I'll post back here soon!

PR #21, opened 2016-12-22
Renamed sdk to client library in README

dan-silver:


dan-silver:
Waiting on #20 to be merged first

PR #26, opened 2017-01-03
Support Node 4 and 5

dan-silver:
- Removed array deconstruction to support older node versions
- Updated Travis to test on Node 4 and 5


PaulStubbs:
Thanks Dan, I tested this and it works in AWS Lambda now.

dan-silver:
Thanks @PaulStubbs for verifying this change! It's been merged to master.

PR #45, opened 2017-06-22
Update README.md

dpim:
Minor wording/emphasis changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,4 +338,4 @@ Please see the [contributing guidelines](CONTRIBUTING.md).
 * [Office Dev Center](http://dev.office.com/)
 
 ## Copyright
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +46,8 @@ All calls to Microsoft Graph are chained together starting with `client.api(path
 * https://graph.microsoft.com/beta/me
 * me/events?$filter=startswith(subject, 'ship')
 
+For more examples of accepted paths, see the [test cases](tests/urlParsing.js).
```
dan-silver: I actually just removed this line since we cover all the patterns here and the test file really isn't a great example. Is this a merge issue or did you want to add it back? (The link is wrong, so we'll have to fix that if keeping)

dpim: Merge issue

PR #46, opened 2017-06-22
Readme updates

dan-silver:


dan-silver:
I'll squash all these down before merging

PR #56, opened 2017-11-15
Export all, remove package.json code dependency

casieber:
1. I've added export statements to index.ts so that consumers can import and major component from the top level of the package rather than needing to import from subdirectories. For example, where previously you would have to do:
    ```
    import { Client } from '@microsoft/microsoft-graph-client';
    import { Options } from '@microsoft/microsoft-graph-client';
    ```
    Now you can do:
    ```
    import { Client, Options } from '@microsoft/microsoft-graph-client';
    ```

2. I've removed the dependency GraphRequest had on package.json and instead added a `PACKAGE_VERSION` constant to the common.ts file that is imported.


lib/src/common.d.ts
```diff
@@ -1,6 +1,7 @@
 export declare let oDataQueryNames: string[];
 export declare const DEFAULT_VERSION = "v1.0";
 export declare const GRAPH_BASE_URL = "https://graph.microsoft.com/";
+export declare const PACKAGE_VERSION = "1.0.0";
```
MIchaelMainer: Added the comment:
// NOTE: The PACKAGE_VERSION version here should match exactly the package version in package.json.

PR #59, opened 2018-01-06
Fix typing for select, expand, and orderby

forivall:
I noticed that the typing for select defines a tuple type with a single item, and not an array type. This caused code such as

```
const userProps: string[] = [
  'city',
  'country'
]

c.api('/me').select(userProps).buildFullUrl()
```

to fail typechecking with the message

```
Argument of type 'string[]' is not assignable to parameter of type 'string | [string]'.
  Type 'string[]' is not assignable to type '[string]'.
    Property '0' is missing in type 'string[]'.
```

This also fixes typechecking for supplying multiple arguments, where `c.api('/me').select('city', 'country').buildFullUrl()` would get the message `Expected 1 arguments, but got 2.`

### DIFF
--- a/spec/core/urlGeneration.ts
+++ b/spec/core/urlGeneration.ts
@@ -32,6 +32,12 @@ cases.push({
 });
 
 cases.push({
+    url: "https://graph.microsoft.com/v1.0/me?$select=displayName,jobTitle",
+    request: client.api("/me")
+                .select("displayName", "jobTitle")
+});
+
+cases.push({
     url: "https://graph.microsoft.com/beta/me?$select=displayName,jobTitle",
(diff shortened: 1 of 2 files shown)


forivall:
Hmm, one thing i thought about: if you want to ensure that arrays have at least one element (i disagree that that should be done, but that's neither here nor there), the workaround that I have to do is

```
const userProps: string[] & [string] = [
  'city',
  'country'
]
```

MIchaelMainer:
@forivall Thank you for taking the time to open this pull request! I won't be able to get to this right away but we'll get this into the next update.

muthurathinam:
Since this is fixed via some other pull request. hence closing.!!

PR #70, opened 2018-05-01
Introduced support to download content as a binary. Minor bug fixes for function signatures

vgrem:
## `.getBinary`

Use `.getBinary()` to stream a download from Microsoft Graph.

```
const fs = require('fs'); // requires filesystem module
client
    .api('/me/drive/root/children/Book.docx/content') // path of  source file in OneDrive
    .getBinary((err, buffer) => {
           fs.writeFileSync('Book.docx', buffer,'binary');
    });
```


muthurathinam:
@vgrem @MIchaelMainer @darrelmiller 
This can be achieved by setting responseType to the get call.

```
client
    .api('/me/drive/root/children/10MB_annual_report_2009.pdf/content')
    .responseType("blob")
    .get()
    .then((res) => {
        console.log("done");
        console.log(res);
    })
    .catch((err) => {
        throw err;
    });
```

Do we really need this api change ? -> https://github.com/microsoftgraph/msgraph-sdk-javascript/pull/70/files#diff-28bf2624bd9644553428f50cb8aac3d7R300

MIchaelMainer:
We don't need it. 

Is it discoverable that someone should needs to set the responseType to "blob"? Do we have docs on it. If it isn't, then getBinary may make more sense. We also have a getStream, does that do the same thing?

muthurathinam:
@MIchaelMainer I agree it is not discoverable as of now. As you mentioned we can update the readme about how to get binary data.
As I am concerned about the size of the output js file, adding such APIs those are achievable in other ways might be unnecessary.
I feel same with the stream APIs, we should have added the guide to achieve that instead of providing new API.

muthurathinam:
@vgrem @MIchaelMainer 

Changes merged with [this PR](https://github.com/microsoftgraph/msgraph-sdk-javascript/pull/96), So closing this one.!!

PR #77, opened 2018-05-23
Fixed Compilation error by changing property value param to accept array of strings in select, expand, orderby methods and by adding required libraries in config.

muthurathinam:


tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "removeComments": true,
     "sourceMap": true,
     "target": "es5",
+    "lib": ["es6", "dom"],
```
MIchaelMainer: What does this change do?

muthurathinam: We were getting compilation error **Cannot find name** for ES6 features like Map, Set, Symbol, WeakMap, WeakSet etc. This was because the compiler could not find a library having these feature set, so I have added *es6* library which has them.
Refer: compiler options [documentation](https://www.typescriptlang.org/docs/handbook/compiler-options.html)

muthurathinam:
I have tested. Shall I merge @MIchaelMainer ?

MIchaelMainer:
Please merge

PR #91, opened 2018-07-30
Append headers instead of setting

muthurathinam:
While adding Authorization header to the request its better to append it


darrelmiller:
@muthurathinam We can't append Authorization header because if one already has been set it will make the HTTP request invalid.

> A sender MUST NOT generate multiple header fields with the same field
>    name in a message unless either the entire field value for that
>    header field is defined as a comma-separated list

For the Authorization header we have two choices, we can either overwrite existing header, or we can do nothing if one already exists.  I'm not sure what the best approach is.

muthurathinam:
I thought, we might require multiple levels of authorization which might requires multiple auth tokens need to be passed. So this appending which creates comma separated tokens for Authorization header will help.

darrelmiller:
@muthurathinam Authorization header is not defined to support multiple tokens.  The Authorization header value is defined as:

    credentials = auth-scheme [ 1*SP ( token68 / #auth-param ) ]

https://tools.ietf.org/html/rfc7235#section-2.1

auth-param can be comma delimited key value pairs if the auth scheme supports it.  However, Graph uses a bearer auth scheme which only supports a single token.

    credentials = "Bearer" 1*SP b64token

https://tools.ietf.org/html/rfc6750#section-2.1

muthurathinam:
@darrelmiller Understood and thanks for the info. 
Am also not sure about the best way for already existing Authorization header, whether to overwrite it or do nothing.!

darrelmiller:
@muthurathinam Let's consider the two options:

Overwrite - This means that if someone "mistakenly" adds an authorization header, then our SDK will overwrite and hopefully do the right thing.  That's assuming they are actually calling a Graph endpoint.

Don't overwrite - This allows someone to manually specify an Authorization header that will be used instead of one generated by the Auth handler.  The risk is someone putting an invalid Authorization header and the request failing.

I'm leaning towards the second option.  If the developer has explicitly put in an Authorization header, then let's assume they are doing the right thing and not touch it.  Once we have 401 handling, if the provided Authorization header is actually invalid, we will get the 401 response and provide a good Auth header value at that point.  So worst case is if someone provides a bad Auth header, we'll do an extra round trip.

muthurathinam:
@darrelmiller 
>So worst case is if someone provides a bad Auth header, we'll do an extra round trip.
Does this extra round trip will happen for every request with which he appends invalid Auth..?

PR #92, opened 2018-08-02
Create two different output js files on with vendor polyfills and one without

muthurathinam:
Two output js files are created
1. With polyfills for fetchapi and es6-promises
2. Without polyfills 

Polyfills for fetchapi and es6-promises are more common, so the host applications might already have the copies of these polyfills. Having ouput without polyfill will actually reduce the sdk file size which fastens the js loading.


muthurathinam:
cc: @lahuey

deepak2016:
@muthurathinam Can you add package, which will get generated after this change, as a comment here?

muthurathinam:
Output files:
1. With polyfills -> [graph-js-sdk-web.js](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/d16fdfce6829a6be9acd3b9a4d8f94a6c76b8917/lib/graph-js-sdk-web.js) [Already present]
2. Without polyfills (FetchAPI and ES6-Promise) -> [graph-js-sdk-core.js](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/d16fdfce6829a6be9acd3b9a4d8f94a6c76b8917/lib/graph-js-sdk-core.js) [Newly generated]

@deepak2016

muthurathinam:
@MIchaelMainer I have updated the [README.md](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/Multiple-Output-JS/README.md#browser) file.

PR #94, opened 2018-08-06
Support for Large file upload task

muthurathinam:
## Large file upload task
In uncertain network connectivity areas uploading a large file would take multiple attempts to get it done. In case of larger files, it would take even more attempts to achieve or even impossible. Large file upload task helps in uploading large files effortlessly.

Implementation is done with the reference of this onedrive's [resumable upload](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/driveitem_createuploadsession).

**To create session and start upload:**

```javascript
let options =  {
    path: "/Documents",
    fileName: file.name,
    rangeSize: (1024 * 1024)
};

MicrosoftGraph.OneDriveLargeFileUploadTask.create(client, file, options)
.then((uploadTask) => {
    uploadTask.upload()
    .then((res) => {
        console.log(res);
        console.log("File Uploaded Successfully.!!");
    })
    .catch((err) => {
        console.log(err);
    });
}).catch((err) => {
    console.log(err);
});
```

**With `async`/`await` syntactic sugar in ECMAScript 2017**

```javascript
async function uploadFile(client, file) {
    try {
        let options = {
            path: "/Documents",
            fileName: file.name,
            rangeSize: (1024 * 1024)
        };
        const uploadTask = await MicrosoftGraph.OneDriveLargeFileUploadTask.create(client, file, options);
        const response = await uploadTask.upload();
        console.log(response);
    } catch (err) {
        console.log(err);
    }
}
```

**We can just resume the broken upload:**

_Lets consider some break down happens in the middle of the uploading, with the uploadTask object in hand resuming comes easy._
```javascript
uploadTask.resume();
```

**Even you can control the whole upload process:**

_Just create the upload task, and play with it by using **sliceFile** and **uploadSlice** methods_

```javascript
let range = uploadTask.getNextRange();
let slicedFile = uploadTask.sliceFile(range);
uploadTask.uploadSlice(slicedFile, range, uploadTask.file.size);
```


spec/tasks/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,206 @@
...
+            nextExpectedRanges: ["327680-"]
+        };
+        uploadTask.updateTaskStatus(statusResponse);
+        let nextRange = uploadTask.getNextRange();
+        assert.equal(nextRange.minValue, 327680);
+        assert.equal(nextRange.maxValue, 328679);
+        done();
+    });
+
+    it('Should return next range as default(empty) range,this is for the upload task completed', (done) => {
```
MIchaelMainer: Nit: space needed 'range, this'

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    // Uploads the file by slicing and uploadin each slice in a sequential manner
+    async upload(): Promise<any> {
+        let self = this;
+        try {
+            while (true) {
+                let nextRange = self.getNextRange();
+                if (nextRange.maxValue === -1) {
+                    throw new Error("Invalid session: Uploading completed");
```
MIchaelMainer: This also aplies to line 181. Would it make sense to create the Error, and then give it a descriptive name so that customers can inspect the error.name and handle it? The error message should give some info on what happed advice on how the error can be handled.

muthurathinam:
@MIchaelMainer Made the changes that you are requested...

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+            self.updateTaskStatus(response);
+            return response;
+        } catch (err) {
+            throw err;
+        }
+    }
+
+    /**
+     * @async
+     * Resume the upload session and continue uploading the file from the place where it left
```
MIchaelMainer: Resume the upload session and continue uploading the file from the last sent range.

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+            minVal = parseInt(firstRange[0]),
+            maxVal = parseInt(firstRange[1]);
+        if (Number.isNaN(maxVal)) {
+            maxVal = this.file.size - 1;
+        }
+        return new Range(minVal, maxVal);
+    }
+
+    /**
+     * To update the exipiration date and the next range
```
MIchaelMainer: nit: expiration

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+        self.uploadSession = uploadSession;
+        self.nextRange = new Range(0, self.options.rangeSize - 1);
+    }
+
+    /**
+     * @static
+     * @async
+     * Creates a LargeFilUploadTask
+     * @param client - The GraphClient instance
+     * @param file - File representated as Blob, File or NodeFile object
```
MIchaelMainer: Change 'NodeFile' to 'Buffer'

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+        options.rangeSize = getValidRangeSize(options.rangeSize);
+        self.options = options;
+        self.uploadSession = uploadSession;
+        self.nextRange = new Range(0, self.options.rangeSize - 1);
+    }
+
+    /**
+     * @static
+     * @async
+     * Creates a LargeFilUploadTask
```
MIchaelMainer: Change LargeFilUploadTask to LargeFileUploadTask

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+        switch(file.constructor.name) {
+            case "Blob":
+                _fileObject.name = (options.fileName !== undefined) ? options.fileName : self.getRandomFileName();
+                _fileObject.content = new File([<Blob>file], _fileObject.name);
+                _fileObject.size = _fileObject.content.size;
+                break;
+            case "File":
+                let _file = <File>file;
+                _fileObject.content = _file;
+                _fileObject.name = _fileObject.name = (options.fileName !== undefined) ? options.fileName : _file.name;
```
MIchaelMainer: Extra '_fileObject.name =' or is this something I don't understand?

Can _file.name ever be undefined? If so, we should use the random name as a last resort.

src/LargeFileUploadTask.ts
```diff
@@ -0,0 +1,338 @@
...
+ * @module LargeFileUploadTask
+ */
+
+import { Client } from "./index";
+import { Range } from "./Range";
+import { getValidRangeSize } from "./LargeFileUploadUtil";
+
+/**
+ * User Options to create the upload task
+ */
```
MIchaelMainer: Consider this comment applicable to all of the files in this PR. Let's add doc comments to all of these types and members. These are all intended to be used by customers so let's make it easier for them to use them. Our .d.ts should be picking these up as well. I'm not sure how we go from .ts to .d.ts. Is there a tool for that? 

/**
 * Interface used to define options when creating an upload task.
 * @interface
 * @property {string} sessionRequestUrl - The default values for parties.
 * @property {string}  fileName               - Optional. Specifies the file name of the file to upload. A file name will be created for you if you do not specify a value.
 * @property {number}  rangeSize          - Optional. Specifies the range chunk size. The default value of ??? is used if a value is not specified. 
 * @property {number} maxTries             - Optional. Specifies the ??? The default value of ??? is used if a value is not specified.
 */

src/LargeFileUploadTask.ts
muthurathinam: >Can _file.name ever be undefined? If so, we should use the random name as a last resort.

_file.name will never be undefined.

src/LargeFileUploadTask.ts
muthurathinam: sure, will define interface in this format.

Typescript compiling does that, when we do compile .ts files using typescript compiler, it generates .js, .d.ts and .js.map files

will investigate and make the compiler take comments from .ts to .d.ts if possible.

muthurathinam:
@MIchaelMainer 
Updated the pull request with the changes:
1. Changed compilation options to take comments from the .ts to .d.ts and .js files
2. Corrected typos in the comments
3. Made file name in options mandatory

.gitignore
```diff
@@ -17,7 +19,8 @@ samples/typescript/graph-typings.d.ts
 samples/typescript/sample.js.map
 src/index.d.ts
 
-spec/types/secrets*
-spec/types/*.js
-spec/types/*.d.ts
-spec/types/*.js.map
\ No newline at end of file
+spec/secrets.ts
```
deepak2016: not required, remove.

muthurathinam:
@darrelmiller @deepak2016 Done changes related to extending LargeFileUploadTask to OneDriveLargeFileUploadTask.

spec/secrets.sample.ts
```diff
@@ -0,0 +1 @@
+export const AccessToken = "";
```
deepak2016: put a comment

spec/tasks/OneDriveLargeFileUploadTask.ts
```diff
@@ -0,0 +1,39 @@
+import { assert } from "chai";
+import { OneDriveLargeFileUploadTask } from "../../lib/src/OneDriveLargeFileUploadTask";
+
+declare const describe, it;
```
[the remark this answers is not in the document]
muthurathinam: Needed... 
>`declare` is used to tell TypeScript that the variable has been created elsewhere. If you use declare, nothing is added to the JavaScript that is generated - it is simply a hint to the compiler.

>For example, if you use an external script that defines var externalModule, you would use declare var externalModule to hint to the TypeScript compiler that externalModule has already been set up

https://stackoverflow.com/a/35020317/1161237

spec/types/OneNote.ts
```diff
@@ -1,6 +1,6 @@
 import { assert } from 'chai'
 
-import { getClient, randomString } from "./test-helper"
+import { getClient, randomString } from "../test-helper"
 import { Notebook, OnenoteSection, OnenotePage } from '@microsoft/microsoft-graph-types-beta'
```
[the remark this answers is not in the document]
muthurathinam: These models are now available in v1.0 so have changed it to use v1.0

.gitignore
muthurathinam: Removed.

spec/secrets.sample.ts
muthurathinam: Commenting done..!!

Issue #97, opened 2018-08-16
Unable to use the library through a proxy


jmorino:
I need to call Graph API through a corporate proxy, but I found no mean to pass options to the underlying fetch() library, in order to specify the proxy URL or a custom HTTP(S) agent.

I think it would be great to provide a way to pass options to fetch(), so it will solve any issue about proxy, timeout or any environment-specific needs.

muthurathinam:
@jmorino Good one. Will implement a way to pass in the options for fetch.

jmorino:
thanks @muthurathinam !

In the meantime, here is my workaround if it can help someone: I just extend the globally-exposed Request class and set a custom HTTPS agent.

```
'use strict';

import { Client } from '@microsoft/microsoft-graph-client'
import HttpsProxyAgent from 'https-proxy-agent'

const OriginalRequest = global.Request;
global.Request = class extends OriginalRequest {
	constructor(...args) {
		super(...args);
		
		const proxy = process.env.HTTPS_PROXY || process.env.https_proxy || null;
		const noproxy = process.env.NO_PROXY || process.env.no_proxy || '';
		const needProxy = proxy && noproxy.split(',').every(h => this.url.indexOf(h.trim()) === -1);
		if (needProxy) {
			this.agent = new HttpsProxyAgent(proxy)
		}
	}
}

export default Client;
```

Issue #39, opened 2017-04-28
Timeout property of superagent library should be exposed


akramhussein:
Our testing experience has shown that intermittently the MSGraph can fail. 

In our test-suite we've extended the time out of Mocha and this has resolved it, but we cannot do this in the main code as timeout property/method of `superagent` is not exposed.

Would it be possible to expose it as an option with a default perhaps?

dan-silver:
That's a great idea! We should support a library default, a client default set in MicrosoftGraph.Client.init, and a per call timeout method.

I can't guarantee when we can add the timeout property, but if you get to it first and submit a pull request I'm happy to review it.

akramhussein:
Hi @dan-silver - sounds good, I will take a look.

PR #105, opened 2018-09-04
Passing options for fetch

muthurathinam:
# Fetch API options
There are node environment specific needs such as passing agent (http(s).Agent instance, allows custom proxy etc.), timeout for request, number of follows for a continuous redirect etc., and these options needs to be passed to fetch call through its RequestInit param (second param).

We have enabled two ways to pass these options for a request.
1. Passing options while initializing the GraphClient. Options can be any [Node specific](https://github.com/bitinn/node-fetch#options) or [Fetch standard](https://fetch.spec.whatwg.org/#requestinit)
Note: This configuration applies to all the requests that are made through GraphClient

```javascript
const client = MicrosoftGraph.init({
    defaultVersion: 'v1.0',
    debugLogging: true,
    authProvider: (done) => {
        done(null, secrets.accessToken);
    },
    fetchOptions: {
        timeout: 5,
        follow: 20
    }
});
```

2. Pass in for every request with `.option` and `.options` functionality.
Note: This applies to the particular request

```javascript
client
    .api('/me')
    .select("displayName")
    .option("redirect", "allow")
    .options({
        timeout: "500"
    })
    .get((err, res) => {
        if (err) {
            console.log(err)
            return;
        }
        console.log(res.displayName);
    });
```
Bugs Fixed:  
1. https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/97
2. https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/39


lib/src/common.d.ts
```diff
@@ -1,16 +1,72 @@
 export declare let oDataQueryNames: string[];
 export declare const DEFAULT_VERSION = "v1.0";
 export declare const GRAPH_BASE_URL = "https://graph.microsoft.com/";
+/**
+ * @constant
+ * A package version
+ * @NOTE: This should be kept up to date with the version used in package.json.
+ * If you are changing this please ensure you are also changing it in package.json.
+ */
```
MIchaelMainer: Comment: I'd like for scenarios to be checked with either a pre-commit hook script or via CI.

src/common.ts
```diff
@@ -2,52 +2,132 @@ export let oDataQueryNames = ["select", "expand", "orderby", "filter", "top", "s
...
+ */
 export const PACKAGE_VERSION = "1.2.0";
 
-// support oData params with and without $ prefix
-oDataQueryNames = oDataQueryNames.concat(oDataQueryNames.map((s) => "$"+s));
+/**
+ * @extends
+ * support oData params with and without $ prefix
+ */
+oDataQueryNames = oDataQueryNames.concat(oDataQueryNames.map((s) => "$" + s));
```
MIchaelMainer: Not part of this PR but we should include search and format as values in oDataQueryNames

README.md
```diff
@@ -247,6 +247,14 @@ You can pass in additional request headers, either individually or in a dictiona
 
 ````
 
+## .option() and .options()
```
MIchaelMainer: Make this a heading 3

lib/src/common.d.ts
muthurathinam: sounds like a very good option. will do that.!

src/common.ts
muthurathinam: Yeah that is something we have in our plate. will make a different PR for that.

PR #116, opened 2018-09-17
Batch Content Redesigned Implementation

muthurathinam:
# Batching
Batching is a way of combining multiple requests to resources in same/different workloads in a single HTTP request. This can be achieved by making a post call with those requests as a JSON payload to $batch endpoint.

### [BatchRequestContent](src/BatchRequestContent.ts)
Component which eases the way of creating batch request payload. This class handles all the batch specific payload construction and stuffs, we just need to worry about individual requests.

```typescript
let folderDetails = {
    "name": "Testing Batch",
    "folder": {}
};
let createFolder = new Request("/me/drive/root/children", {
    method: "POST",
    headers: {
        "Content-type": "application/json"
    },
    body: JSON.stringify(folderDetails)
});

let createFolderStep: BatchRequestStep = {
    id: "1",
    request: createFolder
};
//Create instance by passing a batch request step
let batchContent = new MicrosoftGraph.BatchRequestContent([createFolderStep]);

let fileName = "test.pdf";
let oneDriveFileRequest = new Request(`/me/drive/root:/${fileName}:/content`, {
    method: "GET"
});
let oneDriveFileStep: BatchRequestStep = {
    id: "2",
    request: oneDriveFileRequest,
    dependsOn: ["1"]
};
//Adding a batch request step to the batch
let oneDriveRequestId = batchContent.addRequest(oneDriveFileStep);

//Extracting content from the instance
let content = await batchContent.getContent();

//Making call to $batch end point with the extracted content
let response = await client.api("/$batch").post(content);

```
There are more functionalities like removeRequest, addDependency, removeDependency.

### [BatchResponseContent](src/BatchResponseContent.ts)
Component to simplify the processing of batch responses by providing functionalities like getResponses, getResponseById, getResponsesIterator

```typescript

//Create instance with response from the batch request
let batchResponse = new MicrosoftGraph.BatchResponseContent(response);

//Getting response by id
console.log(batchResponse.getResponseById(oneDriveRequestId));

//Getting all the responses
console.log(batchResponse.getResponses());

//Getting iterator and looping through the responses iterator
let iterator = batchResponse.getResponsesIterator();
let data = iterator.next();
while (!data.done) {
    console.log(data.value[0] + ":" + data.value[1]);
    data = iterator.next();
}

```


src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+        return body;
+    }
+
+    /**
+     * Adds a dependency for a given dependent request
+     *
+     * @param {string} dependentId - The id of the dependent request
+     * @param {string} [dependencyId] - The id of the dependency request, if not specified the preceding request will be considered as a dependency
+     */
+    addDependency(dependentId: string, dependencyId?: string) {
```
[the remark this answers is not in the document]
muthurathinam: [dependencyId] is a way of referring the optional param in doc comment, Array will go like this variableName[ ].

lib/src/BatchRequestContent.d.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param {string} requestId - The id of a request that needs to be removed
+     * @return The boolean indicating removed status
+     */
+    removeRequest(requestId: string): boolean;
+    /**
+     * @async
+     * Serialize content from BatchRequestContent instance
+     * @return The body content to make batch request
+     */
+    content(): Promise<BatchRequestBody>;
```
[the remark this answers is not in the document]
muthurathinam: sounds fine. will change it to getContent().

muthurathinam:
@MIchaelMainer Have done the changes that you have requested. Please have a look.

lib/src/BatchRequestContent.d.ts
```diff
@@ -0,0 +1,127 @@
...
+     * Constructs a BatchRequestContent instance
+     * @param {BatchRequestStep[]} [requests] - Array of requests value
+     */
+    constructor(requests?: BatchRequestStep[]);
+    /**
+     * Adds a request to the batch request content
+     * @param {BatchRequestStep} request - The request value
+     * @return The id of the added request
+     */
+    addRequest(request: BatchRequestStep): string;
```
pankajpalOlk: Do we need to return the id of the added Request? Could it be a 'void' method..

src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+            }
+        }
+    }
+
+    /**
+     * Adds a request to the batch request content
+     * @param {BatchRequestStep} request - The request value
+     * @return The id of the added request
+     */
+    addRequest(request: BatchRequestStep): string {
```
pankajpalOlk: rename to requestStep

src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+        }
+        if (self.requests.size === limit) {
+            let error = new Error(`Maximum requests limit exceeded, Max allowed number of requests are ${limit}`);
+            error.name = "Limit Exceeded Error";
+            throw error;
+        }
+        if (self.requests.has(request.id)) {
+            let error = new Error(`Adding request with duplicate id ${request.id}, Make the id of the requests unique`);
+            error.name = "Duplicate RequestId Error";
+            throw error;
```
pankajpalOlk: so what happens at the caller if we throw from here..?

src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+            headers[key] = value;
+        });
+        if (Object.keys(headers).length) {
+            requestData.headers = headers;
+        }
+        if (request.method === RequestMethod.PATCH || request.method === RequestMethod.POST || request.method === RequestMethod.PUT) {
+            requestData.body = await BatchRequestContent.getRequestBody(request);
+        }
+        /**
+         * Check any other property needs to be used from the Request object and add them
```
pankajpalOlk: TODO: Check any other ...

it helps in textual searching for TODOs later on.

src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+        try {
+            let cloneReq = request.clone();
+            body = await cloneReq.json();
+            bodyParsed = true;
+        } catch (e) {
+
+        }
+        if (!bodyParsed) {
+            try {
+                if (typeof Blob !== "undefined") {
```
pankajpalOlk: What is this Blob ? some global ?

src/BatchRequestContent.ts
```diff
@@ -0,0 +1,460 @@
...
+                    let blob = await request.blob();
+                    let reader = new FileReader();
+                    body = await new Promise(resolve => {
+                        reader.addEventListener("load", function () {
+                            let dataURL = <string>reader.result,
+                                regex = new RegExp("^\s*data:(.+?\/.+?(;.+?\=.+?)*)?(;base64)?,(.*)\s*$"),
```
pankajpalOlk: It is generally a good idea to add a couple of examples in the comments for regular expression patterns.

src/BatchRequestContent.ts
muthurathinam: Yes, Blob is a global object present only in the browser environment. Since code is being used both by node and browsers, we have this check

lib/src/BatchRequestContent.d.ts
muthurathinam: Instead of returning nothing, returning id gives the sense that the request is been added. This is why we are returning id.

PR #122, opened 2018-09-28
feat(polyfill): do not polyfill

### LINKED ISSUE #113: avoid polyfill fetch and promise
Polyfills should be in user land

https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/src/GraphRequest.ts#L2

can we remove those and let the user polyfill them?

sibelius:
Do not polyfill Promise, neither fetch
Let users bring their own implementations

fix #113


sibelius:
tks

PR #144, opened 2019-02-28
Implementation of AuthenticationHandlerOptions and AuthenticationProviderOptions

muthurathinam:
## Summary

* Implemented AuthenticationHandlerOptions
* Implemented AuthenticationProviderOptions
* Added isComplete() to PageIterator
* Added validation for consistency between package version in package.json and Constants.ts
* Added tasks test in launch.json
* Updated dependencies to latest version in package.json
* Fixed delay calculation bug on exponential backoff in RetryHandler
* Fixed bug that is not returning proper error for graph error

## Test plan

Tests added to the current test suite, running `npm test` should not fail on any test cases.

## Types of changes

## Checklist

-   [x] I have updated the documentation accordingly.


src/GraphResponseHandler.ts
```diff
@@ -49,20 +49,24 @@ export class GraphResponseHandler {
...
+						}
+					});
 				});
-			});
-		} else {
-			return Promise.resolve(rawResponse.body);
+			} else {
+				return Promise.resolve(rawResponse.body);
+			}
+		} catch (error) {
```
MIchaelMainer: We can omit try-catch-throw since we aren't inspecting the error object. Did you intend to inspect this error? same with [R130](https://github.com/microsoftgraph/msgraph-sdk-javascript/pull/144/files#diff-097803c7ff7847cd97efb185745e148dR130)

muthurathinam: Yes, with line number 130 I did inspect the error object. I have added this for the code clarity to bascially understand that this will throw error.

PR #161, opened 2019-04-11
Adding latest ES outputted files

muthurathinam:
## Summary

Added ES folder which contains js files with latest es and created an output js file named graph-es-sdk.js which also has the latest es

## Motivation

* To support the advantages that comes with the latest es version
* To leverage the advantages that comes with the module based build systems like rollup and webpack

## Types of changes

## Checklist


tsconfig-es.json
```diff
@@ -0,0 +1,17 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
```
nmetulev: Should be es6 or esnext so es modules are generated

muthurathinam: Yeah will fix that...

muthurathinam: fixed

src/index.ts
```diff
@@ -5,11 +5,7 @@
  * -------------------------------------------------------------------------------------------
  */
 
-import { init } from "./PolyFill";
-/**
- * Polyfilling Promise and fetch if not present
- */
-init();
+import "./PolyFill";
```
nmetulev: I recommend not including the polyfils for modules as they are usualy unnecessary. This is causing a weird error for me when running in Chrome and removing this line fixes it.

tsconfig-es.json
nmetulev: Thank you!

rollup.config.js
```diff
@@ -0,0 +1,15 @@
+import { terser } from "rollup-plugin-terser";
```
nmetulev: I'd argue that you don't need the `graph-es-sdk.js` file as the other bundled js files work just as fine in the browsers. A single file with all modules defeats the purpose of modules.

tsconfig-es.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "./tsconfig-base.json",
+	"compilerOptions": {
+		"module": "ESNext",
```
nmetulev: some language features from esnext are missing in Edge and this is causing syntax errors in that environment. Please use `es6` for both the `module` and `target` properties

src/index.ts
muthurathinam: Not including polyfill will be a breaking change, since this makes user to include his own polyfill in case if its not natively present. We will eventually remove this in some time.

tsconfig-es.json
nmetulev: I got it working on my end by using babel to transpile to something that works on edge too. This is not currently a blocking issue but it would have made things easier.

PR #164, opened 2019-04-23
Telemetry Handler implementation.

muthurathinam:
## Summary

Implemented Telemetry Handler to facilitate the feature usage metrics.

## Test plan

`npm run test`

## Types of changes

## Checklist

### DIFF
--- a/src/HTTPClientFactory.ts
+++ b/src/HTTPClientFactory.ts
@@ -40,6 +40,12 @@ export class HTTPClientFactory {
 	 * Creates HTTPClient with default middleware chain
 	 * @param {AuthenticationProvider} authProvider - The authentication provider instance
 	 * @returns A HTTPClient instance
+	 *
+	 * NOTE: These are the things that we need to remember while doing modifications in the below default pipeline.
+	 * 		* HTTPMessageHander should be the last one in the middleware pipeline, because this makes the actual network call of the request
+	 * 		* TelemetryHandler should be the one prior to the last middleware in the chain, because this is the one which actually collects and appends the usage flag and placing this handler 	*		  before making the actual network call ensures that the usage of all features are recorded in the flag.
+	 * 		* The best place for AuthenticationHandler is in the starting of the pipeline, because every other handler might have to work for multiple times for a request but the auth token for
+	 * 		  them will remain same. For example, Retry and Redirect handlers might be working multiple times for a request based on the response but their auth token would remain same.
 	 */
 	public static createWithAuthenticationProvider(authProvider: AuthenticationProvider): HTTPClient {
 		const authenticationHandler = new AuthenticationHandler(authProvider);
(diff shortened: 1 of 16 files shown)


src/HTTPClientFactory.ts
```diff
@@ -43,15 +44,18 @@ export class HTTPClientFactory {
 	public static createWithAuthenticationProvider(authProvider: AuthenticationProvider): HTTPClient {
 		const authenticationHandler = new AuthenticationHandler(authProvider);
 		const retryHandler = new RetryHandler(new RetryHandlerOptions());
+		const telemetryHandler = new TelemetryHandler();
 		const httpMessageHandler = new HTTPMessageHandler();
+
```
MIchaelMainer: I suggest that you comment about the importance of the order of these handlers and how someone else in the future should update this pipeline of handlers.

src/middleware/options/TelemetryHandlerOptions.ts
```diff
@@ -0,0 +1,86 @@
...
+import { MiddlewareOptions } from "./IMiddlewareOptions";
+
+/**
+ * @enum
+ * @property {number} NONE - The hexadecimal flag value for nothing enabled
+ * @property {number} REDIRECT_HANDLER_ENABLED - The hexadecimal flag value for redirect handler enabled
+ * @property {number} RETRY_HANDLER_ENABLED - The hexadecimal flag value for retry handler enabled
+ * @property {number} AUTHENTICATION_HANDLER_ENABLED - The hexadecimal flag value for the authentication handler enabled
+ */
+export enum FeatureUsageFlag {
```
MIchaelMainer: It was my understanding that the handler pipeline should be [customizable per spec](https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/middleware/middleware.md). We should add a CUSTOM_HANDLER_ENABLED flag when that work happens. Would we update IMiddleWare to require it set it at that point?

src/HTTPClientFactory.ts
muthurathinam: will be more helpful, will do...

src/middleware/options/TelemetryHandlerOptions.ts
muthurathinam: we can figure out this from the flag value right ? let say I constructed a custom middleware which uses our Redirect handler and my own custom middlewares. in such case the feature usage flag will be set only for redirect handler. then we can conclude that the application is using custom middleware. your thoughts ?

MIchaelMainer: I just realized that I missed the createWithMiddleware in the client factory. This comment is not actionable in this PR and requires greater discussion and update to spec.

The current flag values can't be use to determine whether there is a custom pipeline or custom handlers. For example, let's say a customer creates a pipeline with our retry, redirect, telemetry, and auth handler, in addition to a custom handler. The flags would look like our default pipeline. 

What my real concern here is that as we start to see adoption of our client versions with the middleware, we are starting to see issues pop-up. One issue involved the handler order. We will see issues when customers start making changes to default handler order. We may want to investigate ways to enforce known handler order. Perhaps a middleware factory or arranger to aid customers in the creation of a pipeline - it'd have business rules for ordering default handlers. 

Another concern I have is with a custom handlers. It would be convenient to know whether a request was made with a custom handler when working with customers on issues. While customers may tell us in an issue or incident, sometimes it will be expedient to look up the logs. Knowing that a custom handler exists will help us pinpoint debugging efforts.

muthurathinam: Yeah, I completely agree with you. This will help a lot.

PR #185, opened 2019-06-07
Updated additional resources links

Dingenis:
## Summary
The links pointed to projects that are archived and have been replaced by different ones. Updated the links to the new projects.

## Motivation

When I was looking into getting started with this sdk, I found out that some links were pointing to old archived projects which have been replaced. By changing the links the users won't be redirected to the old archived project but to the up to date one. 

## Test plan

See if links are properly updated by clicking.

## Types of changes

-   [x] Bugfix, link updates.


muthurathinam:
@Dingenis Good job finding the breakage and Thank you for the PR.

PR #219, opened 2019-08-16
Adds support for multiple cached authorities for MSAL auth provider

bkrabach:
Adds support for multiple cached authorities for MSAL auth provider by explicitly passing the msalApplication.authority value in the tokenRequest.

If a user is logged in with multiple MSAL authorities on the same site, there are multiple tokens and authorities cached on the client.  Without explicitly passing the authority, the MSAL application is unable to determine which token to retrieve and fails with: Multiple authorities found in the cache. Pass authority in the API overload.

We encountered this scenario when requiring users to sign into a Sharepoint account, but also sign into a separate AAD/MSA account for MS Graph access.  This update fixes the issue.

### DIFF
--- a/src/ImplicitMSALAuthenticationProvider.ts
+++ b/src/ImplicitMSALAuthenticationProvider.ts
@@ -70,6 +70,7 @@ export class ImplicitMSALAuthenticationProvider implements AuthenticationProvide
 		}
 		if (this.msalApplication.getAccount()) {
 			const tokenRequest: AuthenticationParameters = {
+				authority: this.msalApplication.authority,
 				scopes,
 			};
 			try {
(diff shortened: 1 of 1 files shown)


darrelmiller:
@bkrabach Thank you for this PR.  My only concern is that we ran into an issue with this approach in the MSAL.NET library.  When the authority ends with /common /organizations or /consumers it causes and extra refresh token request.  https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1456   Are you able to verify if your fix works when the client application has been configured to work for multiple organizations?

bkrabach:
@darrelmiller The scenario we're using it with right now is that we allow users to log in with a combination of MSA+Sharepoint AAD or Microsoft AAD + Sharepoint AAD.  All three use different authorities, however none of them end in /common, /organizations, or /consumers - they all end in their respective tenant id's.  This fix solved the problem for us and lets us use the various combinations that we needed.

nikithauc:
@bkrabach  Thank you for contributing! The ImplicitMsalProvider will be deprecated in the upcoming release 3.0.0. No further changes will be made to ImplicitMsalProvider.

PR #227, opened 2019-09-13
made testing Handler

krototype:
## Summary

Testing Handler - Helps in mocking Microsoft Graph to generate all kinds of responses.
Currently supports:

Manual Mode:
-- Client Level
-- Request Level

Random Mode

I have to add unit tests for the Testing Handler.

## Motivation

Helps the developers to check their software with hard to generate responses as well.

## Types of changes

-   [X] New feature (non-breaking change which adds functionality)

## Checklist

-   [ ] I have read the **CONTRIBUTING** document.


src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+	/**
+	 * @private
+	 * Generates responseHeader
+	 * @param {number} statusCode - the status code to be returned for the request
+	 * @param {string} statusMessage - the status message to be returned for the request
+	 * @param {string} requestID - request id
+	 * @param {Date} requestDate - date of the request
+	 * @returns response Header
+	 */
+	private createResponseHeader(statusCode: number, statusMessage: string, requestID: string, requestDate: Date) {
```
darrelmiller: I think this should be createResponseHeaders

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+	 * @public
+	 * @async
+	 * To execute the current middleware
+	 * @param {Context} context - The context object of the request
+	 * @returns A Promise that resolves to nothing
+	 */
+	public async execute(context: Context): Promise<void> {
+		try {
+			const testingHandlerOptions = this.getOptions(context);
+			context.response = this.setStatusCode(testingHandlerOptions, context.request as string, context.options.method as string);
```
MIchaelMainer: I suggest that we alter createResponse to take an HttpMethod, so that we could call this.createResponse to get a Response object instead of calling this.setStatusCode to get a Response object. 

Then have createResponse call setStatusCode before forming the Response object.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+	 * @param {string} requestMethod - the API method for the request
+	 * @returns Response object
+	 */
+	private setStatusCode(testingHandlerOptions: TestingHandlerOptions, requestURL: string, requestMethod: string): Response {
+		try {
+			testingHandlerOptions.statusMessage = "Some error happened";
+			if (testingHandlerOptions.testingStrategy === TestingStrategy.MANUAL) {
+				if (testingHandlerOptions.statusCode === undefined) {
+					try {
+						const urlMethod = this.getRelativeURL(new RegExp("http(s)://graph.microsoft.com/[^/]*", "g"), requestURL);
```
MIchaelMainer: Change `urlMethod` to `relativeUrl`.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+	/**
+	 * @private
+	 * To get the options for execution of the middleware
+	 * @param {Context} context - The context object
+	 * @returns A options for middleware execution
+	 */
+	private getOptions(context: Context): TestingHandlerOptions {
+		let options: TestingHandlerOptions;
+		if (context.middlewareControl instanceof MiddlewareControl) {
+			options = context.middlewareControl.getMiddlewareOptions(TestingHandlerOptions) as TestingHandlerOptions;
```
bilalgce: Add null checks for the input parameter & middleware control

krototype:
> I was expecting to find a spec here: https://github.com/microsoftgraph/msgraph-sdk-design/tree/master/middleware. Can you add the spec you have in OneNote to the design repo so everyone can reference and review it? You've captured a lot of quality ideas. It will be useful to leverage across SDKs, customers can provide input, it is easier to reference from our PRs, and someone else could benefit from what you've captured.

https://github.com/microsoftgraph/msgraph-sdk-design/pull/22
I have created a PR for this spec document.

src/middleware/TestingHandler.ts
krototype: We have set Default cases and some error handling, so it seems we won't need null checks here.

src/middleware/TestingHandler.ts
MIchaelMainer: Please provide a response about why this should stay as is before resolving the conversation with no action.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+					}
+				} else {
+					if (requestURL === this.redirectURL) {
+						const statusCode: number = testingHandlerOptions.statusCode;
+						if (statusCode === 301 || statusCode === 302 || statusCode === 303 || statusCode === 307 || statusCode === 308) {
+							testingHandlerOptions.statusCode = 404;
+						}
+					}
+				}
+			} else if (testingHandlerOptions.testingStrategy === TestingStrategy.RANDOM) {
```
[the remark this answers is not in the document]
MIchaelMainer: Please provide a response about why this should stay as is before resolving the conversation with no action.

src/middleware/TestingHandler.ts
MIchaelMainer: Ugh, was looking at out dated version. All good.

src/middleware/TestingHandler.ts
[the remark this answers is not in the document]
MIchaelMainer: Ugh, was looking at out dated version. All good.

src/middleware/options/TestingHandlerOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+ * -------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.
+ * See License in the project root for license information.
+ * -------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @module TestingHandlerOptions
+ */
+
```
muthurathinam: Remove unused imports in the file

src/middleware/options/TestingHandlerOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+import { FetchOptions } from "../../IFetchOptions";
+
+import { MiddlewareOptions } from "./IMiddlewareOptions";
+import { methodStatusCode } from "./TestingHandlerData";
+import { TestingStrategy } from "./TestingStrategy";
+
+export class TestingHandlerOptions implements MiddlewareOptions {
```
muthurathinam: JSDoc comments are missing in this file.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+export class TestingHandler implements Middleware {
+	/**
+	 * @private
+	 * A member holding options to customize the handler behavior
+	 */
+
+	private options: TestingHandlerOptions;
+
+	public statusCode: number;
```
muthurathinam: I don't find the use of this variable. where it is used ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	public statusMessage: string;
```
muthurathinam: I don't find the use of this variable. where it is used ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	// the mode of testing handler
+	public testingStrategy: TestingStrategy;
```
muthurathinam: I don't find the use of this variable. where it is used ? the same is defined under testingHandlerOptions, do we need it here ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	// container for the manual map that has been written by the client
+	public manualMap: Map<string, Map<string, number>>;
```
muthurathinam: Does it have to be public variable ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	// dummy url that we are using
+	public redirectURL: string = "https://dummylocation.microsoft.com";
```
muthurathinam: Does it have to be public variable ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	/**
+	 * @private
+	 * Generates responseHeader
+	 * @param {number} statusCode - the status code to be returned for the request
+	 * @param {string} statusMessage - the status message to be returned for the request
+	 * @param {string} requestID - request id
+	 * @param {Date} requestDate - date of the request
+	 * @returns response Header
+	 */
+	private createResponseHeaders(statusCode: number, statusMessage: string, requestID: string, requestDate: Date) {
```
muthurathinam: Unused argument statusMessage. I think we should remove this.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+		// creates a responseHeader based on the status code
+		const responseHeader: any = {};
```
muthurathinam: Why the type of the responseHeader is any ? can we use Header type ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	 * @param {number} statusCode - the status code to be returned for the request
+	 * @param {string} statusMessage - the status message to be returned for the request
+	 * @param {string} requestID - request id
+	 * @param {Date} errDate - date of the request
+	 * @returns response body
+	 */
+	private createResponseBody(statusCode: number, statusMessage: string, requestID: string, requestDate: Date) {
+		// response body gets created as empty for passed cases
+		// response body contains error field for failure scenarios
+		let responseBody;
```
muthurathinam: declare the explicit type for the variable.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	 * @private
+	 * creates a response object out of responseHeader and responseBody
+	 * @param {TestingHandlerOptions} testingHandlerOptions - The TestingHandlerOptions object
+	 * @param {string} requestURL - the URL for the request
+	 * @returns Response object
+	 */
+	private createResponse(testingHandlerOptions: TestingHandlerOptions, requestURL: string, requestMethod: string): Response {
+		try {
+			// creates a response Object out of responseHeader and responseBody
+			let responseBody;
```
muthurathinam: Type declaration needed for the variable responseBody and responseHeader.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	 * To fetch the status code from the map(if needed), then returns response by calling createResponse
+	 * @param {TestingHandlerOptions} testingHandlerOptions - The TestingHandlerOptions object
+	 * @param {string} requestURL - the URL for the request
+	 * @param {string} requestMethod - the API method for the request
+	 * @returns Response object
+	 */
+	private setStatusCode(testingHandlerOptions: TestingHandlerOptions, requestURL: string, requestMethod: string) {
+		try {
+			// setting some random status code
+			testingHandlerOptions.statusMessage = "Some error happened";
```
muthurathinam: If we are hard coding the default error message, we can do the same inside TesingHandlerOptions constructor or may be we can set this as a default value for the TestingHandlerOption.statusMessage.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	/**
+	 * @private
+	 * To fetch the relative URL out of the complete URL
+	 * @param {RegExp} pattern - the regex pattern for the URL
+	 * @param {string} urlMethod - the complete URL
+	 * @returns the relative URL
+	 */
+	private getRelativeURL(pattern: RegExp, urlMethod: string): string {
```
muthurathinam: If the purpose of the method is to return relative url from the absolute url then it makes sense to have the regex pattern inside the function.

urlMethod needs to be changed to url.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+		try {
+			// setting some random status code
+			testingHandlerOptions.statusMessage = "Some error happened";
+			if (testingHandlerOptions.testingStrategy === TestingStrategy.MANUAL) {
+				// inside the Manual mode
+				if (testingHandlerOptions.statusCode === undefined) {
+					// manual mode with no status code, can be a global level or request level without statusCode
+					try {
+						// chacking Manual Map for exact match
+						const urlMethod = this.getRelativeURL(new RegExp("http(s)://graph.microsoft.com/[^/]*", "g"), requestURL);
```
muthurathinam: The regex has an unnecessary grouping i.e http(s) here the grouping **(s)**.

Is it meant to match http and https ?

If the request url is `https://graph.microsoft.com/v1.0/users?$filter=startswith(givenName,'J')`  then in this case the relative url returned would be `/users?$filter=startswith(givenName,'J')`. Do we expect the same ?
or do we need to strip out the query params too ?

In the former case user has to provide the manualMap with all possible query params that would be unnecessary.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+
+	/**
+	 * @private
+	 * Fetches a random status code for the RANDOM mode from the predefined array
+	 * @param {string} requestMethod - the API method for the request
+	 * @returns the random status code
+	 */
+	private getRandomStatusCode(requestMethod: string): number {
```
muthurathinam: change request method type from string to RequestMethod which is defined in the file RequestMethod.ts

muthurathinam: this applies to all the places where the requestMethod is used

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+			testingHandlerOptions.statusMessage = "Some error happened";
+			if (testingHandlerOptions.testingStrategy === TestingStrategy.MANUAL) {
+				// inside the Manual mode
+				if (testingHandlerOptions.statusCode === undefined) {
+					// manual mode with no status code, can be a global level or request level without statusCode
+					try {
+						// chacking Manual Map for exact match
+						const urlMethod = this.getRelativeURL(new RegExp("http(s)://graph.microsoft.com/[^/]*", "g"), requestURL);
+						testingHandlerOptions.statusCode = this.manualMap.get(urlMethod).get(requestMethod);
+					} catch (error) {
```
muthurathinam: when does the catch block will get executed ? I believe it won't. Did you test this part of the code ?

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+	private setStatusCode(testingHandlerOptions: TestingHandlerOptions, requestURL: string, requestMethod: string) {
+		try {
+			// setting some random status code
+			testingHandlerOptions.statusMessage = "Some error happened";
+			if (testingHandlerOptions.testingStrategy === TestingStrategy.MANUAL) {
+				// inside the Manual mode
+				if (testingHandlerOptions.statusCode === undefined) {
+					// manual mode with no status code, can be a global level or request level without statusCode
+					try {
+						// chacking Manual Map for exact match
```
muthurathinam: typo in the comment chacking -> checking

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,265 @@
...
+						this.manualMap.forEach((value: Map<string, number>, key: string) => {
+							const regexUrl = new RegExp(key);
+							if (regexUrl.test(urlMethod)) {
+								testingHandlerOptions.statusCode = this.manualMap.get(key).get(requestMethod);
+							}
+						});
+
+						// Case of redirection or request url not in map
+						if (testingHandlerOptions.statusCode === undefined) {
+							if (requestURL === this.redirectURL) {
```
muthurathinam: what if the user provided requestURL is exactly same as the redirectURL that you have ? we need to generate randomized url for the redirecturl so that we can minimize the collision.

src/middleware/TestingHandler.ts
muthurathinam: we should not rely on the hardcoded url. it is possible that user can set his base url as the one you provided. so we need to randomize this in order to reduce the possibility of the collision.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,271 @@
...
+	 * @param {number} statusCode - the status code to be returned for the request
+	 * @param {string} requestID - request id
+	 * @param {string} requestDate - date of the request
+	 * @returns response Header
+	 */
+	private createResponseHeaders(statusCode: number, requestID: string, requestDate: string) {
+		const responseHeader: Headers = new Headers();
+
+		responseHeader.append("Cache-Control", "private");
+		responseHeader.append("Transfer-Encoding", "");
```
darrelmiller: Remove this header.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,271 @@
...
+	 * @private
+	 * @param {number} statusCode - the status code to be returned for the request
+	 * @param {string} requestID - request id
+	 * @param {string} requestDate - date of the request
+	 * @returns response Header
+	 */
+	private createResponseHeaders(statusCode: number, requestID: string, requestDate: string) {
+		const responseHeader: Headers = new Headers();
+
+		responseHeader.append("Cache-Control", "private");
```
darrelmiller: Set the Cache-Control header to no-store.  I don't see a reason why we would want any downstream cache to store this response and serve it up.  The response isn't coming over the wire so there is no performance gain to allow caching.

krototype:
Once the reviewers approve the PR, I would be squashing all the commits

src/middleware/TestingHandler.ts
```diff
@@ -191,28 +193,35 @@ export class TestingHandler implements Middleware {
...
 
-						// Case of redirection or request url not in map
-						if (testingHandlerOptions.statusCode === undefined) {
-							if (requestURL === this.redirectURL) {
-								// we send a 404 after a single redirect if it's done (chain contains redirect Handler)
-								testingHandlerOptions.statusCode = 404;
-							} else {
-								throw new Error("API not available in map");
-							}
+					// Case of redirection or request url not in map
```
muthurathinam: Shouldn't this come as a first if statement in the previous if else chain ? just an optimization.

krototype:
@darrelmiller @MIchaelMainer have a look in the changes so that I can merge.

darrelmiller:
@krototype  While not specifically related to TestingHandler I would like to see a solution that allows the user to not have to rebuild the default middleware chain in order to insert one custom handler.   In .NET we did this by adding a two methods to the factory, one to return an ordered array of the default handlers and one to turn the array into a chain.  https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dev/src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs#L122   This allows a user to get the array of default handlers, manipulate it in some way and then wire them together.  

Requiring the developer to recreate the chain to add one handler means that if we add a new piece of middleware to the chain in a future release any dev that has customized the chain at all will not get the new piece of middleware.

Can we investigate if this pattern we used in C# will work in JavaScript also?  Thanks.

src/middleware/TestingHandler.ts
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	/**
+	 * To execute the current middleware
+	 * @public
+	 * @async
+	 * @param {Context} context - The context object of the request
+	 * @returns A Promise that resolves to nothing
+	 */
+	public async execute(context: Context): Promise<void> {
```
darrelmiller: We should not be returning failure conditions for every request.  For random we should only return failures on a small percentage of requests.  That percentage should be configurable.  For manual configuration we should only return errors if the request URL maps to one of the failure cases.

The goal of this handler is that a developer can actually work with the application and can verify that the application can recover from failures.  This would not work if every response was a failure.

krototype: @darrelmiller 
- Do we need to increase the priority of the success scenarios for the Random case?
- Do we need to give users the option of configuring the type of requests in Random mode?
- For the Manual mode, we return an error only when map points to one of the failure scenarios (we even consider 3xx case as an error scenario, in js-SDK and hence error objects are returned).

darrelmiller: @krototype We shouldn't return success status codes.  We should pass the request through the testing handler to the Graph API and allow the regular request to pass through.  The idea of the testing handler is to allow a developer to use their application but have random errors be injected into their application so that they can see whether the application handles server generated failures.  I realize now that when we had the conversation about what status code were returned that I got confused sidetracked away from the original goal.  The only status codes that we should return are ones are difficult to manually trigger.  e.g. 5XX and 429.  The rest of the error codes like 4XX and 3XX are unnecessary to ever return in random mode.  I need to investigate more regarding manual mode because I don't see how 3XX could ever be considered an error.

docs/TestingHandlerSamples.md
```diff
@@ -0,0 +1,116 @@
...
+//Declaring a Manual Map for all the requests
+const manualMap: Map<string, Map<string, number>> = new Map([
+["/me/messages/.*", new Map([["GET", 200], ["PATCH", 200]])],
+["/users/me/sendMail", new Map([["GET", 404], ["POST", 307]])]
+]);
+
+
+//Declaring a Custom Middleware Chain
+const telemetryHandler = new TelemetryHandler();
+const testingHandler = new TestingHandler(new TestingHandlerOptions(TestingStrategy.MANUAL), manualMap);
```
darrelmiller: Rename to ChaosHandler

src/middleware/ChaosHandler.ts
```diff
@@ -0,0 +1,289 @@
...
+	private async createResponse(chaosHandlerOptions: ChaosHandlerOptions, context: Context): Promise<void> {
+		try {
+			let responseBody: any;
+			let responseHeader: Headers;
+			let requestID: string;
+			let requestDate: Date;
+			const requestURL = context.request as string;
+			const requestMethod = context.options.method as RequestMethod;
+
+			const passThrough = this.setStatusCode(chaosHandlerOptions, requestURL, requestMethod);
```
muthurathinam: it doesn't make sense to return passthrough value from setStatusCode method. better have the pass through logic in a separate function.

src/middleware/ChaosHandler.ts
```diff
@@ -0,0 +1,289 @@
...
+			if (passThrough === false) {
```
muthurathinam: `if (passThrough === false)` can be simplified to `if (!passThrough)`

src/middleware/ChaosHandler.ts
```diff
@@ -185,7 +197,7 @@ export class ChaosHandler implements Middleware {
 	 * @param {string} requestURL - the URL for the request
 	 * @param {string} requestMethod - the API method for the request
 	 */
-	private setStatusCode(chaosHandlerOptions: ChaosHandlerOptions, requestURL: string, requestMethod: RequestMethod): boolean {
+	private getStatusCode(chaosHandlerOptions: ChaosHandlerOptions, requestURL: string, requestMethod: RequestMethod) {
```
muthurathinam: name getStatusCode means this function would return some status code... instead of updating it in the chaosHandlerOptions just return the status code from the function and assign the same in the callee function

muthurathinam: or let the name be setStatusCode.

PR #290, opened 2020-06-09
Update README

behnammodi:
Fix miss MSALAuthenticationProviderOptions import


ddyett:
looking at adding both classes to the index.ts to avoid having to add the lib portion but will update the documentation if we do.

Issue #291, opened 2020-06-09
URLs containing escaped or unescaped equals characters are truncated or extraneous text added


markwahl-msft:
**Describe the bug**
A partner reports that the new Graph Explorer fails to submit the following GET request:

https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/88b35805-6815-4334-ae57-ef2010e4c855?$expand=accessPackageAssignmentPolicies,accessPackageResourceRoleScopes($expand=accessPackageResourceRole,accessPackageResourceScope)

It seems to only request the following url, truncating the end:

https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/88b35805-6815-4334-ae57-ef2010e4c855?$expand=accessPackageAssignmentPolicies,accessPackageResourceRoleScopes($expand=

The url works with other tools.

When attempting to use escape characters, I observed an even more surprising behavior, when I put into the URI field "%3D" in place of the "=" character (I changed the GUID to that from one or my tenants), then what it sent had a spurious "=undefined" at the end?

https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/09378296-4cc6-4ce2-ab73-94fd2c4fdaff?$expand%3DaccessPackageAssignmentPolicies,accessPackageResourceRoleScopes($expand%3DaccessPackageResourceRole,accessPackageResourceScope)=undefined

**To Reproduce**
1. Go to https://developer.microsoft.com/en-us/graph/graph-explorer and sign in as an administrator user in a tenant where entitlement management is configured. 
2. Provide in the URI field the string

https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/?$expand=accessPackageAssignmentPolicies,accessPackageResourceRoleScopes($expand=accessPackageResourceRole,accessPackageResourceScope)

and click "Run query". 
4. See error of bad request returned.  In modern Edge, F12 shows that the request that was sent had
been truncated to https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/?$expand=accessPackageAssignmentPolicies,accessPackageResourceRoleScopes($expand

**Expected behavior**

The URI that is entered by the user is sent to the service.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]  Windows 10
 - Browser [e.g. chrome, safari]  Edge
[...]

PR #306, opened 2020-07-28
Issue 291 - Changed parse path function considering nested query parameters.

nikithauc:
[URL parsing failed ](https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/291)  because the parsePath() function did not considering nested query functions. Changed the function to split the query parameters on the first equals sign([ OData Parsing rules](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360925)).

Also, added comments to few query option functions mentioning their expected behaviour.

### DIFF
--- a/spec/core/urlGeneration.ts
+++ b/spec/core/urlGeneration.ts
@@ -113,6 +113,37 @@ cases.push({
 		.query({ $select: "id" }),
 });
 
+// handling an invalid input
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$select=displayName,title&select=value&test",
+	request: client
+		.api("/me/people")
+		.version("v1.0")
+		.select(["displayName", "title"])
+		.query({ select: "value" })
+		.query("test"),
+});
+
+// handling an invalid input
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$expand=address($select=home,$expand=city)&$select=home,displayName,title&select=value&test",
+	request: client
+		.api("/me/people?$expand=address($select=home,$expand=city)&$select=home")
+		.version("v1.0")
+		.select(["displayName", "title"])
+		.query({ select: "value" })
+		.query("test"),
+});
+
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$expand=home($select=home)&name=test",
+	request: client.api("/me/people").query("?name=test&$expand=home($select=home)"),
+});
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$expand=home($select=home)&name=test",
+	request: client.api("/me/people?name=test&$expand=home($select=home)"),
+});
+
 cases.push({
 	url: "https://graph.microsoft.com/v1.0/me/drive/root?$expand=children($select=name),permissions",
 	request: client
--- a/spec/core/urlGeneration.ts
+++ b/spec/core/urlGeneration.ts
@@ -92,6 +92,28 @@ cases.push({
 });
 
 cases.push({
+	url: "https://graph.microsoft.com/beta/me/people?$select=displayName,title,id&$count=false&$expand=a($expand=a,b)",
+	request: client
+		.api("/me/people")
+		.version("beta")
+		.select(["displayName", "title"])
+		.count(true)
+		.expand("a($expand=a,b)")
+		.query("$select=id")
+		.query("$count=false"),
+});
+
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$select=displayName,title,id&select=value",
+	request: client
+		.api("/me/people")
+		.version("v1.0")
+		.select(["displayName", "title"])
+		.query({ select: "value" })
+		.query({ $select: "id" }),
+});
+
+cases.push({
 	url: "https://graph.microsoft.com/v1.0/me/drive/root?$expand=children($select=name),permissions",
 	request: client
 		.api("me/drive/root")
--- a/src/GraphRequest.ts
+++ b/src/GraphRequest.ts
@@ -170,14 +170,7 @@ export class GraphRequest {
 			// Capture query string into oDataQueryParams and otherURLQueryParams
 			const queryParams = path.substring(queryStrPos + 1, path.length).split("&");
 			for (const queryParam of queryParams) {
-				const qParams = queryParam.split("=");
-				const key = qParams[0];
-				const value = qParams[1];
-				if (oDataQueryNames.indexOf(key) !== -1) {
-					this.urlComponents.oDataQueryParams[key] = value;
-				} else {
-					this.urlComponents.otherURLQueryParams[key] = value;
-				}
+				this.query(queryParam);
 			}
 		}
 	};
@@ -430,7 +423,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add query string for filter OData Query param
+	 * To add query string for filter OData Query param. The request URL accepts only one $filter Odata Query option and it's value is set to the most recently passed filter query string.
 	 * @param {string} filterStr - The filter query string
 	 * @returns The same GraphRequest instance that is being called with
 	 */
@@ -441,7 +434,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add criterion for search OData Query param
+	 * To add criterion for search OData Query param. The request URL accepts only one $search Odata Query option and it's value is set to the most recently passed search criterion string.
 	 * @param {string} searchStr - The search criterion string
 	 * @returns The same GraphRequest instance that is being called with
 	 */
@@ -452,7 +445,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add number for top OData Query param
+	 * To add number for top OData Query param. The request URL accepts only one $top Odata Query option and it's value is set to the most recently passed number value.
 	 * @param {number} n - The number value
 	 * @returns The same GraphRequest instance that is being called with
 	 */
@@ -463,7 +456,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add number for skip OData Query param
+	 * To add number for skip OData Query param. The request URL accepts only one $skip Odata Query option and it's value is set to the most recently passed number value.
 	 * @param {number} n - The number value
 	 * @returns The same GraphRequest instance that is being called with
 	 */
@@ -474,7 +467,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add token string for skipToken OData Query param
+	 * To add token string for skipToken OData Query param. The request URL accepts only one $skipToken Odata Query option and it's value is set to the most recently passed token value.
 	 * @param {string} token - The token value
 	 * @returns The same GraphRequest instance that is being called with
 	 */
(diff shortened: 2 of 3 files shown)


src/GraphRequest.ts
```diff
@@ -501,19 +494,26 @@ export class GraphRequest {
...
 			for (const key in queryDictionaryOrString) {
 				if (queryDictionaryOrString.hasOwnProperty(key)) {
-					otherURLQueryParams[key] = queryDictionaryOrString[key];
+					paramKey = key;
+					paramValue = queryDictionaryOrString[key];
 				}
 			}
 		}
+		if (oDataQueryNames.indexOf(paramKey) !== -1) {
+			this.urlComponents.oDataQueryParams[paramKey] = paramKey === "$expand" || paramKey === "$select" || paramKey === "$orderby" ? (this.urlComponents.oDataQueryParams[paramKey] ? this.urlComponents.oDataQueryParams[paramKey] + "," + paramValue : paramValue) : paramValue;
```
zengin, line 513: oDataQueryParams were not being modified in the context of a `query` call, now they are. This seems to be an unexpected side effect.

I see that it was being modified in the `parsePath` function before, but this is a public function and you are changing the behavior not only for `parsePath` code path, but for all public callers.

src/GraphRequest.ts
```diff
@@ -430,7 +423,7 @@ export class GraphRequest {
 
 	/**
 	 * @public
-	 * To add query string for filter OData Query param
+	 * To add query string for filter OData Query param. The request URL accepts only one $filter Odata Query option and it's value is set to the most recently passed filter query string.
```
zengin, line 426: nit: typo `it's` -> `its` in this and similar comments below

src/GraphRequest.ts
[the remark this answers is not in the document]
nikithauc: The existing code contains a [addCsvQueryParameter function](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/870477b4eefd5ba7c5838cbd64280c5cc2cbc362/src/GraphRequest.ts#L193)  which is called from the .select, expand and orderby query functions. The function does a comma separated concatenation for multiple values of a given key. The idea to introduce this in the query and thereby the parsePath function is to maintain consistency. For example, if we use client.api("/me?$select=id").select("name") then the query parameter part of the output would be  "/me?$select= id,name". But client.api("/me?$select=id&$select=name") would output the query parameter part as "/me?$select=name" that is replace the first query param value. Therefore, I added the comma separated concatenation from the addCsvQueryParameter to query function so that different functions to build a query string behave the same way.  

We do support  parsing "http://host/service/Products?$select=Name,Description&$expand=Category($select=Name)". The query parameter is separated on the first equals and nested part is maintained in the value part as is. I ran test case and got the expected output ->  

>"/Products$select=Name,Description&$expand=Category($select=Name)":"https://graph.microsoft.com/v1.0/Products?$select=Name,Description&$expand=Category($select=Name)" .

nikithauc: As explained in the previous comment, the idea is to maintain consistency. The reason to call the query function from parsePath function is to avoid duplicate comment.

src/GraphRequest.ts
```diff
@@ -501,19 +494,26 @@ export class GraphRequest {
...
+		let paramValue: string | number;
 		if (typeof queryDictionaryOrString === "string") {
-			const querySplit = queryDictionaryOrString.split("=");
-			const queryKey = querySplit[0];
-			const queryValue = querySplit[1];
-			otherURLQueryParams[queryKey] = queryValue;
+			const indexOfFirstEquals = queryDictionaryOrString.indexOf("="); // The query key-value pair must be split on the first equals sign to avoid error while parsing nested query parameters
+			const qParams = [queryDictionaryOrString.substring(0, indexOfFirstEquals), queryDictionaryOrString.substring(indexOfFirstEquals + 1, queryDictionaryOrString.length)];
+			paramKey = qParams[0];
+			paramValue = qParams[1];
```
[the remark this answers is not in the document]
nikithauc: You are right. I will change that.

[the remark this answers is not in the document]
nikithauc: Fixed this.

src/GraphRequest.ts
nikithauc: Fixed the spelling error

src/GraphRequest.ts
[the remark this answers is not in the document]
nikithauc: I broke the ternary conditions into separate lines. The linting wouldn't let me commit the code with if else loop. Please check now.

nikithauc: @ddyett @MIchaelMainer Would like to know your thoughts here please.

src/GraphRequest.ts
```diff
@@ -501,19 +494,29 @@ export class GraphRequest {
...
+		let paramKey: string;
+		let paramValue: string | number;
 		if (typeof queryDictionaryOrString === "string") {
-			const querySplit = queryDictionaryOrString.split("=");
-			const queryKey = querySplit[0];
-			const queryValue = querySplit[1];
-			otherURLQueryParams[queryKey] = queryValue;
+			/* The query key-value pair must be split on the first equals sign to avoid errors in parsing nested query parameters.
+			 Example-> "/me?$expand=home($select=city)" */
+			const indexOfFirstEquals = queryDictionaryOrString.indexOf("=");
```
zengin, line 502: Sorry, I didn't catch this in the first iteration, but we need further validation of queryDictionaryOrString. 

If the string is `test` without any equals sign, indexOfFirstEquals becomes `-1`, `paramKey` becomes empty string, `paramValue` becomes `test`.

src/GraphRequest.ts
```diff
@@ -501,19 +494,29 @@ export class GraphRequest {
...
 				if (queryDictionaryOrString.hasOwnProperty(key)) {
-					otherURLQueryParams[key] = queryDictionaryOrString[key];
+					paramKey = key;
+					paramValue = queryDictionaryOrString[key];
 				}
 			}
 		}
+		if (oDataQueryNames.indexOf(paramKey) !== -1) {
+			const currentValue = this.urlComponents.oDataQueryParams[paramKey];
+			const isValueAppendable = (paramKey === "$expand" || paramKey === "$select" || paramKey === "$orderby") && currentValue;
```
zengin, line 515: small perf: if `currentValue` comes first, this can be short circuited for cases where `currentValue` is false.

```suggestion
			const isValueAppendable = currentValue && (paramKey === "$expand" || paramKey === "$select" || paramKey === "$orderby");
```

src/GraphRequest.ts
MIchaelMainer: Sounds like we need a private `parseQueryString`. 

Both `parsePath` and `query` functions need to parse a querystring and apply the odata and non-odata query parameters to `urlComponents` object. The `query` function had the same bug as `parsePath` where it split on  `=` without considering nested queries. 

The change to parsePath should be treated as public since it is always used to parse the path argument supplied to `client.api("")`.

spec/core/urlGeneration.ts
```diff
@@ -91,6 +91,28 @@ cases.push({
...
+		.count(true)
+		.expand("a($expand=a,b)")
+		.query("$select=id")
+		.query("$count=false"),
+});
+
+cases.push({
+	url: "https://graph.microsoft.com/v1.0/me/people?$select=displayName,title,id&select=value",
+	request: client
+		.api("/me/people")
```
MIchaelMainer, line 109: Let's add cases where nested queries are set in `.api()` as that will exercise `parsePath`.

MIchaelMainer, line 109: We need test cases where nested queries are set in `.api("")` so that we exercise `parsePath` and parsing the querystring

MIchaelMainer: Ugh, I thought this wasn't saved. Okay, resolved this dup

jobala:
These changes work for me, except for the edge case pointed by @zengin.

I have a few suggestions regarding improving the code's readability.

1. We can improve readability of the if block by using variable names instead of expressions. 
```
const hasOdataQueryName = oDataQueryNames.indexOf(paramKey) !== -1;

if (hasOdataQueryName) {
...
}
```
2. By extracting smaller functions from query() ie `handleQueryString()` & `handleQueryDictionary` we make it easy to see what the function is doing.

What do you think of extracting the changes added to query() in a separate function then call the extracted function in parsePath() and query()?

src/GraphRequest.ts
```diff
@@ -501,19 +494,29 @@ export class GraphRequest {
...
+			 Example-> "/me?$expand=home($select=city)" */
+			const indexOfFirstEquals = queryDictionaryOrString.indexOf("=");
+			paramKey = queryDictionaryOrString.substring(0, indexOfFirstEquals);
+			paramValue = queryDictionaryOrString.substring(indexOfFirstEquals + 1, queryDictionaryOrString.length);
 		} else {
 			for (const key in queryDictionaryOrString) {
 				if (queryDictionaryOrString.hasOwnProperty(key)) {
-					otherURLQueryParams[key] = queryDictionaryOrString[key];
+					paramKey = key;
+					paramValue = queryDictionaryOrString[key];
```
ddyett, line 509: shouldn't there be a break here? if not can you add a comment about the intent of the code here.  once paramKey is set it's not clear why continue

src/GraphRequest.ts
```diff
@@ -501,19 +494,29 @@ export class GraphRequest {
 	 * @returns The same GraphRequest instance that is being called with
```
ddyett, line 494: can you comment here on what the modification the graphrequest is.  It currently reads like I passed in something and got the same thing back.

src/GraphRequest.ts
nikithauc: Handled both the cases mentioned above. 
1. Checking if the first character is ? and cropping it off.
2. If the query parameter is not a valid string or a not a valid parameter appending it the query parameter string. 
Example -> client.api(/me).query("?test").query(123) will be built as "/me?test&123"/. The idea is to not change or replace any user input (even if invalid) and forward the request to the API.

spec/core/urlGeneration.ts
nikithauc: Added. Please verify.

src/GraphRequest.ts
nikithauc: It was a mistake on my part. Corrected the code so that the iteration records key and value for each parameter in the dictionary.

nikithauc:
> These changes work for me, except for the edge case pointed by @zengin.
> 
> I have a few suggestions regarding improving the code's readability.
> 
> 1. We can improve readability of the if block by using variable names instead of expressions.
> 
> ```
> const hasOdataQueryName = oDataQueryNames.indexOf(paramKey) !== -1;
> 
> if (hasOdataQueryName) {
> ...
> }
> ```
> 
> 1. By extracting smaller functions from query() ie `handleQueryString()` & `handleQueryDictionary` we make it easy to see what the function is doing.
> 
> What do you think of extracting the changes added to query() in a separate function then call the extracted function in parsePath() and query()? 

I have added a private function which is called by parsePath() and query() function. I have tried using variables in place of expressions wherever the expression appears more than once. And tried breaking down to small functions.

spec/core/urlGeneration.ts
nikithauc: Added more test cases. Please verify. The urlParsing.ts file has more test cases which apply to the parsePath() function.

src/GraphRequest.ts
nikithauc: Added more information in some comments. There are more comments like this one. Will change them in another PR.

Issue #298, opened 2020-07-16
Making a call that uses  application/xhtml+xml fails


thewahome:
# Bug Report
## Prerequisites

-   [x] Can you reproduce the problem?
-   [ ] Are you running the latest version?
-   [x] Are you reporting to the correct repository?
-   [ ] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

## Description

I noticed that all calls by the sdk use the `application/json ` header. I am making a request to `https://graph.microsoft.com/v1.0/me/onenote/sections/{section-id}/pages`  
and it requires `application/xhtml+xml` as the request header. 

The payload is: 

```
<!DOCTYPE html>
<html>
  <head>
    <title>A page with a block of HTML</title>
  </head>
  <body>
    <p>This page contains some <i>formatted</i> <b>text</b>.</p>
  </body>
</html>
```

Looking at the network call, the content-type is set to `application/json` and not the one passed into the client.headers() function

**Console Errors:** [Is there any console error]

**Screenshots:** [If applicable, add screenshots to help explain your problem]

## Steps to Reproduce

1. Log in to graph explorer
2. Run the create page query after setting the required section id
3. notice the bad request and the response as the screenshot above shows

**Expected behavior:** 
A page in my one note would be created

**Actual behavior:**
I receive a bad request error

## Additional Context

Add any other context about the problem here..

## Usage Information

SDK Version - [SDK version you are using]

-   [ ] Node (Check, if using Node version of SDK)

> Node Version - [The version of Node you are using]

-   [ ] Browser (Check, if using Browser version of SDK)

> Browser Name - [The name of Browser that you are using for SDK]

> Version - [The version of the browser you are using]

Issue #293, opened 2020-06-25
Empty post requests from Graph Explorer fail in the SDK


ddyett:
Related to https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/555

PR #309, opened 2020-08-05
Js issue 293 298 emptybody content type

nikithauc:
Made changes for the following issues 
1. https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/298
2. https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/293

Added a private function which checks if the header contains the 'content-type', if not sets the default 'content-type' as 'application/json'. Also, added a null check to the serializeContent function.


spec/development/workload/OneNote.ts
```diff
@@ -85,6 +85,22 @@ describe("OneNote", function() {
 		}
 	});
 
+	it("Create a OneNote page with application/xhtml+xml page content", async () => {
+		try {
```
zengin, line 89: Please remove the try catch block if you are going to rethrow the same exact error in the catch block without additional handling.

src/GraphRequest.ts
```diff
@@ -302,6 +303,29 @@ export class GraphRequest {
 		}
 	}
 
+	/**
+	 * @private
+	 * Checks if the content-type is present in the _headers property. If not present, defaults the content-type to application/json
```
zengin, line 308: Below code doesn't set the `Content-Type` if `_headers` is defined but there is no `Content-Type` key.

src/GraphRequest.ts
```diff
@@ -302,6 +303,29 @@ export class GraphRequest {
...
+	 * @param none
+	 * @returns nothing
+	 */
+	private setHeaderContentType(): void {
+		if (this._headers === undefined || this._headers === null) {
```
zengin, line 313: if (!this._headers)

src/GraphRequest.ts
```diff
@@ -302,6 +303,29 @@ export class GraphRequest {
...
+			this.header("Content-Type", "application/json");
+		}
```
zengin, line 315:
```suggestion
			return;
		}
```

src/GraphRequest.ts
```diff
@@ -550,13 +574,15 @@ export class GraphRequest {
...
 			method: RequestMethod.POST,
 			body: serializeContent(content),
-			headers:
-				typeof FormData !== "undefined" && content instanceof FormData
-					? {}
-					: {
-							"Content-Type": "application/json",
-					  },
 		};
+		const className: string = content === undefined || content === null ? undefined : content.constructor.name;
```
zengin, line 578:
```suggestion
		const className = content?.constructor?.name;
```

src/GraphRequestUtil.ts
```diff
@@ -41,7 +41,7 @@ export const urlJoin = (urlSegments: string[]): string => {
  */
 
 export const serializeContent = (content: any): any => {
-	const className: string = content === undefined ? undefined : content.constructor.name;
+	const className: string = content === undefined || content === null ? undefined : content.constructor.name;
```
zengin, line 44:
```suggestion
	const className = content?.constructor?.name;
```

src/GraphRequest.ts
jobala: Curious to know why an empty return is better than no return @zengin

zengin: If `headers` is `undefined` or `null`, current code will set the header and continue to the logic that follows, which attempts to iterate over `headers` to search for `content-type`.

jobala: Interesting, this makes lines 316 to 326 not necessary, right?

Since by line 316 the content-type property will always be present.

src/GraphRequestUtil.ts
nikithauc: I tried using this but the build failed. Will look at it in a separate PR.

zengin: interesting, please share your findings when you get a chance. It could be that the TypeScript version we are using does not support this feature.

nikithauc: Yeah, I suspect the same reason for the failure. I will let you know the reason.

spec/development/workload/OneNote.ts
nikithauc: Removed the case. There are more tests in the project which are using try catch and throwing the error. I will clean that up in a separate PR.

src/GraphRequest.ts
zengin: Not for all cases, that's why we need a return to stop execution if the headers are null or undefined. If they are defined, we can still proceed with the logic to check if "content-type" header exists.

More reading on this topic:
1. Arrowhead anti pattern and early return as a solution: https://stackoverflow.com/questions/17804005/how-to-prevent-the-arrowhead-anti-pattern
2. Early return vs single exit: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

Early returns are a widely adapted style in these situations. Example:
```
if (someObject == null)
{
  return false;
}

// someObject can safely be assumed not null at this point:
var a = someObject.SomeProperty;
```

src/GraphRequest.ts
```diff
@@ -302,6 +303,30 @@ export class GraphRequest {
...
+	private setHeaderContentType(): void {
+		if (!this._headers) {
+			this.header("Content-Type", "application/json");
+			return;
+		}
+		let isContentTypePresent = false;
+		const headerKeys = Object.keys(this._headers);
+		for (const headerKey of headerKeys) {
+			if (headerKey.toLowerCase() === "content-type") {
+				isContentTypePresent = true;
```
zengin, line 321: @jobala, another opportunity for early return is at line 321.

Without defining `isContentTypePresent`, we can return directly here if "content-type" is found. Then, reaching line 325 will mean that the loop has finished without finding "content-type" (otherwise we would have hit the early return case), so we can skip the if check there and blindly assign "application/json".

So semantically equivalent code starting from 317:
```
		const headerKeys = Object.keys(this._headers);
		for (const headerKey of headerKeys) {
			if (headerKey.toLowerCase() === "content-type") {
				return; // content-type found, no update needed
			}
		}
		// Default the content-type to application/json in case the content-type is not present in the header
		this.header("Content-Type", "application/json");
```

@nikithauc, please either change the function logic to what I have above if it makes sense to you or add a break statement after setting `isContentTypePresent`, so that we don't continue iterating over headers after finding the searched header.

nikithauc: Yes. That's right. Will make this change

nikithauc: Added the change.

package-lock.json
```diff
@@ -1085,7 +1085,8 @@
 		"asynckit": {
 			"version": "0.4.0",
 			"resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
-			"integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k="
+			"integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k=",
+			"dev": true
```
zengin, line 1089: Do we need the development versions of these packages?

nikithauc: That got auto updated when I did an npm install.

zengin: Looks like these packages won't be installed if the `--production` flag is used. Not sure if this is what we want. See the last response here:
https://stackoverflow.com/questions/49809490/what-is-dev-true-in-package-lock-json-for

Do you mind testing the changes after removing these flags? Sorry, I am not well-versed in npm, trying to make sure this won't break production behavior.

tsconfig-es.json
```diff
@@ -1,10 +1,11 @@
 {
 	"extends": "./tsconfig-base.json",
 	"compilerOptions": {
-		"module": "es6",
+		"module": "CommonJS",
```
MIchaelMainer, line 4: Why are we making this change? Will this be a breaking change?

nikithauc: This change is made because the build failed when trying to import the  form-data module. This worked along with esModuleInterop being set true.

src/GraphRequest.ts
```diff
@@ -9,6 +9,8 @@
  * @module GraphRequest
  */
 
+import FormData from "form-data";
```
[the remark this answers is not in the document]
nikithauc: I added this line because the [test cases](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/ed902ecf040d5f2ef644447d6754fb423ac33e0e/spec/development/workload/OneNote.ts#L72)   which used form-data failed because this import wasn't present. I will test this change for a browser sample. But I think the import is needed.

tsconfig-es.json
MIchaelMainer: I suspect that we don't want to make this change. We have a tsconfig for ES6 and CommonJS. I don't think we want the `import FormData from "form-data";` in GraphRequest. FormData is available in most browsers. It will be up to the customer to import formdata when running on Node. You should add an update to the readme to provide instuction that they need to install and import  formdata if they are running on Node and they need to make calls to Microsoft Graph that accept formdata.

src/GraphRequest.ts
```diff
@@ -632,13 +654,15 @@ export class GraphRequest {
...
 			body: serializeContent(content),
-			headers:
-				typeof FormData !== "undefined" && content instanceof FormData
-					? {}
-					: {
-							"Content-Type": "application/json",
-					  },
 		};
+		const className: string = content === undefined || content === null ? undefined : content.constructor.name;
+		if (typeof FormData !== "undefined" && className === "FormData") {
```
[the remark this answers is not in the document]
nikithauc: I tried sending a content of type FormData through the samples/node application  as follows-> 
`const FormData =require('form-data');
const formData = new FormData();
			formData.append("Presentation", fs.createReadStream("./spec/sample_files/onenotepage_fileattachment.html"));
			formData.append("fileBlock1", fs.createReadStream("./sample.png"));
client.api(`/me/onenote/sections/1-b3076c16-9e2d-4391-a4ca-061142aeac24/pages`).headers({"User":"Name"}).post(formData).then((res) => {
    console.log(res);
})
.catch((err) => {
    console.log(err);
});`
Removed the import of form-data from the GraphRequest.ts imports. I think we need to explicitly import form-data in the GraphRequest.ts file. But will have to look more into the browser issues.

package-lock.json
nikithauc: The dev=true is set because the npm packages have been added as dev dependencies. I noticed that the flag got auto-data when I added form-data as a dev dependency in package.json. I will look into your suggestion.

nikithauc: Made changes that no longer require the form-data package to be installed.

src/GraphRequest.ts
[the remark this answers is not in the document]
nikithauc: Removed the import

tsconfig-es.json
nikithauc: Restored the tsconfig files

PR #313, opened 2020-08-14
Task/parse path with regex

jobala:
## Summary
Uses regex  to get host, version, path and query string from url instead of traversing the url string.

## Motivation

The motivation for this change is to improve how we get url properties.

## Test plan

Run `npm t`

## Types of changes

## Checklist

### DIFF
--- a/src/GraphRequestUtil.ts
+++ b/src/GraphRequestUtil.ts
@@ -58,3 +58,99 @@ export const serializeContent = (content: any): any => {
 	}
 	return content;
 };
+
+export class PathUtils {
+	private static hostRe = /([a-z]+.microsoft.[a-z]+)/g;
+	private static versionRe = /(beta|v1.0)/g;
+	private static fullPathRe = /((v1.0|beta).*?(?=\?|$))/g;
+	private static shortPathRe = /(.*?)((?=\?|$))/g;
+	private static queryRe = /\?.*/g;
+
+	private static getPathFromFullPath(url: string): string {
+		const version = PathUtils.getVersionFrom(url);
+		const path = url.match(PathUtils.fullPathRe);
+		if (!path) {
+			return;
+		}
+
+		return path.shift().replace(version, "");
+	}
+
+	private static getPathFromShortPath(url: string) {
+		const path = url.match(PathUtils.shortPathRe);
+
+		if (!path) {
+			return;
+		}
+
+		return path.shift();
+	}
+
+	/**
+	 * Matches the host from url. Given then url https://graph.microsoft.com/v1.0/me/events
+	 * it will return graph.microsoft.com.
+	 *
+	 * @param url
+	 * @returns host
+	 */
+	public static getHostFrom(url: string): string {
+		const host = url.match(PathUtils.hostRe);
+
+		if (!host) {
+			return "graph.microsoft.com";
+		}
+
+		return host.shift();
+	}
+
+	/**
+	 * Matches the version from url. Given the url https://graph.microsoft.com/v1.0/me/events,
+	 * it will return the version, v1.0 or beta.
+	 *
+	 * @param url
+	 * @returns version
+	 */
+	public static getVersionFrom(url: string): string {
+		const version = url.match(PathUtils.versionRe);
+
+		if (!version) {
+			return "v1.0";
+		}
+
+		return version.shift();
+	}
+
+	/**
+	 * Matches the path from url. If the user passes a full url like
+	 * https://graph.microsoft.com/v1.0/me/events it will return v1.0/me
+	 * And when the users passes a short url like me/events it will return me/events
+	 *
+	 * @param url
+	 * @returns path
+	 */
+	public static getPathFrom(url: string) {
+		const isFullPath: boolean = url.includes("microsoft");
+
+		if (isFullPath) {
+			return PathUtils.getPathFromFullPath(url);
+		} else {
+			return PathUtils.getPathFromShortPath(url);
+		}
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


src/GraphRequestUtil.ts
```diff
@@ -58,3 +58,99 @@ export const serializeContent = (content: any): any => {
 	}
 	return content;
 };
+
+export class PathUtils {
+	private static hostRe = /([a-z]+.microsoft.[a-z]+)/g;
+	private static versionRe = /(beta|v1.0)/g;
```
zengin, line 64: Q: how do we guarantee that false positives won't appear in other parts of the url? For example:
`https://graph.microsoft.com/v1.0/me/messages?$filter=startswith(subject, 'anothermatchingurl.microsoft.com')`
or
`https://graph.microsoft.com/v1.0/me/messages?$filter=startswith(subject, 'beta')`

zengin, line 64: in `(beta|v1.0)` expression, `.` is a wildcard character, so it will match things like: `v1a0` and `v1b0` etc. The expression should be `(beta|v1\.0)`.

src/GraphRequestUtil.ts
```diff
@@ -58,3 +58,99 @@ export const serializeContent = (content: any): any => {
...
+	/**
+	 * Matches the path from url. If the user passes a full url like
+	 * https://graph.microsoft.com/v1.0/me/events it will return v1.0/me
+	 * And when the users passes a short url like me/events it will return me/events
+	 *
+	 * @param {string} url
+	 * @returns {string} path
+	 */
+	public static getPathFrom(url: string) {
+		const isFullPath: boolean = url.includes("microsoft");
```
zengin, line 132: False positive example where `isFullPath` becomes `true`:
`/me/messages?$filter=startswith(subject, 'microsoft')`

src/GraphRequestUtil.ts
```diff
@@ -58,3 +58,99 @@ export const serializeContent = (content: any): any => {
...
+		const version = PathUtils.getVersionFrom(url);
+		const path = url.match(PathUtils.fullPathRe);
+		if (!path) {
+			return;
+		}
+
+		return path.shift().replace(version, "");
+	}
+
+	private static getPathFromShortPath(url: string) {
```
zengin, line 79: I would propose the following refactor for code reuse.

It also handles strong typing as current functions have empty returns vs string returns. I think empty returns in TS become `undefined` if the return type is `string`, but returning a defaultValue might be better for the consumers of the API. Even if we want to return `undefined`, it is better to be explicit about it.
```
	private static matchRegex(url:string, regex: RegExp, defaultValue: string = "", clearAfterMatch : string = null) : string {
		const matched = url.match(regex);
		if (!matched) {
			return defaultValue;
		}

		if (clearAfterMatch) {
			return matched.shift().replace(clearAfterMatch, "");
		}

		return matched.shift();
	}

	private static getPathFromFullPath(url: string): string {
		const version = this.getVersionFrom(url);
		return this.matchRegex(url, this.fullPathRe, "", version);
	}

	private static getPathFromShortPath(url: string) {
		return this.matchRegex(url, this.shortPathRe);
	}

	public static getHostFrom(url: string): string {
		return this.matchRegex(url, this.hostRe, "graph.microsoft.com");
	}

	public static getVersionFrom(url: string): string {
		return this.matchRegex(url, this.versionRe, "v1.0");
	}

	public static getQueryStringFrom(url: string): string {
		return this.matchRegex(url, this.queryRe, "", "?");
	}
```

src/GraphRequestUtil.ts
```diff
@@ -58,3 +58,99 @@ export const serializeContent = (content: any): any => {
...
+
+	/**
+	 * Matches the version from url. Given the url https://graph.microsoft.com/v1.0/me/events,
+	 * it will return the version, v1.0 or beta.
+	 *
+	 * @param url
+	 * @returns {string} version
+	 */
+	public static getVersionFrom(url: string): string {
+		const version = url.match(PathUtils.versionRe);
```
nikithauc, line 114: If a user enters an invalid url. For example -> A simple spelling mistake such as `client.api("https://graph.microsoft.com/bta/me").get()`. In this case the match fails and returns "v1.0 " as a version. It would not be a clear indicator of the error. 
It would be easier to simply forward the incorrect url and let the user see the error when the response fails.  The other advantage of doing this would be that we wouldn't be blocking a valid url which has a different syntax but is marked as invalid by the code.

src/GraphRequestUtil.ts
nikithauc, line 64: What do you think about not hardcoding the versions? My concern is that we shouldn't miss updating this when we have newer versions.

src/GraphRequestUtil.ts
zengin: I agree that an approach that is based on url segments rather than values would work better, given that the OData URLs are also mostly based on path segments.

src/GraphRequestUtil.ts
jobala: This will match multiple groups for example `graph.microsoft.com` and `anothermatchingurl.microsoft.com` but the getHostMethod will return the first match which is `graph.microsoft.com`

src/GraphRequestUtil.ts
jobala: @nikithauc what do you think of throwing an error if there's no matching version before the user makes a request?

src/GraphRequestUtil.ts
jobala: I think this is fine as the versions don't change often.

jobala:
Hi @nikithauc & @zengin, thanks for taking your time to review this PR.

I am closing this PR because it tries to fix something that is not broken.

PR #325, opened 2020-09-15
Update OneDriveLargeFileUploadTask.ts

hasridha:
A similar PR may already be submitted! Please search among the [Pull request](https://github.com/microsoftgraph/msgraph-sdk-javascript/pulls) before creating one.

Thanks for submitting a pull request! Please provide enough information so that others can review your pull request.

**NOTE: PR's will be accepted only in case of appropriate information is provided below**

## Summary

Encoding the entire URL does not encode URL components correctly. When the SDK's client tries to encode these URL components in the parameters on the call to OneDriveLargeFileUploadTask.create, then the resultant file in OneDrive has the URL encoded filename (as opposed to the user-friendly/decoded filename). Therefore, fix the encoding in the SDK's URL components itself, directly.

## Motivation

Explain the **motivation** for making this change. What existing problem does the pull request solve?

Trying to upload files with the '#' or '%' character in the filename ends with network errors of the form -
400 status response "The parameter item does not exist in method getByPath" 

## Test plan

## Closing issues

Fixes #

## Types of changes

-   [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

-   [X] I have read the **CONTRIBUTING** document.
-   [X] My code follows the code style of this project.

### DIFF
--- a/src/tasks/OneDriveLargeFileUploadTask.ts
+++ b/src/tasks/OneDriveLargeFileUploadTask.ts
@@ -59,7 +59,7 @@ export class OneDriveLargeFileUploadTask extends LargeFileUploadTask {
 		if (path[path.length - 1] !== "/") {
 			path = `${path}/`;
 		}
-		return encodeURI(`/me/drive/root:${path}${fileName}:/createUploadSession`);
+		return `/me/drive/root:${path.split('/').map(p => encodeURIComponent(p)).join('/')}${encodeURIComponent(filename)}:/createUploadSession`;
 	}
 
 	/**
--- a/src/tasks/OneDriveLargeFileUploadTask.ts
+++ b/src/tasks/OneDriveLargeFileUploadTask.ts
@@ -59,7 +59,7 @@ export class OneDriveLargeFileUploadTask extends LargeFileUploadTask {
 		if (path[path.length - 1] !== "/") {
 			path = `${path}/`;
 		}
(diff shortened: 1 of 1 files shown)


nikithauc (approved):

src/tasks/OneDriveLargeFileUploadTask.ts
```diff
@@ -59,7 +59,7 @@ export class OneDriveLargeFileUploadTask extends LargeFileUploadTask {
 		if (path[path.length - 1] !== "/") {
 			path = `${path}/`;
 		}
-		return encodeURI(`/me/drive/root:${path}${fileName}:/createUploadSession`);
+		return `/me/drive/root:${path.split('/').map(p => encodeURIComponent(p)).join('/')}${encodeURIComponent(fileName)}:/createUploadSession`;
```
zengin, line 62: would be nice to have a comment here in case someone attempts to revert it back to the original for optimization.
```suggestion
		// we choose to encode each component of the file path separately because when encoding full URI
		// with encodeURI, special characters like # in the file name doesn't get encoded as desired
		return `/me/drive/root:${path.split('/').map(p => encodeURIComponent(p)).join('/')}${encodeURIComponent(fileName)}:/createUploadSession`;
```

zengin (approved):

zengin (approved):

ddyett (approved):

PR #331, opened 2020-09-30
Release 2.1.0

nikithauc:
Bumping the version to 2.1.0 for changes in [2.1.0-Preview.1](https://github.com/microsoftgraph/msgraph-sdk-javascript/releases/tag/2.1.0-Preview.1)


src/GraphRequestUtil.ts
```diff
@@ -41,7 +41,7 @@ export const urlJoin = (urlSegments: string[]): string => {
  */
 
 export const serializeContent = (content: any): any => {
-	const className: string = content.constructor.name;
+	const className: string = content === undefined ? undefined : content.constructor.name;
```
baywet, line 44:
```suggestion
	const className: string = content && content.constructor && content.constructor.name;
```

src/HTTPClient.ts
```diff
@@ -33,6 +33,37 @@ export class HTTPClient {
...
+	 * @public
+	 * To get an array of Middleware, used in middleware chain
+	 * @returns An array of middlewares
+	 */
+	public getMiddlewareArray(): Middleware[] {
+		const middlewareArray: Middleware[] = [];
+		let currentMiddleware = this.middleware;
+		while (currentMiddleware) {
+			middlewareArray.push(currentMiddleware);
+			if (typeof currentMiddleware.getNext !== "undefined") {
```
baywet, line 46:
```suggestion
			if (currentMiddleware.getNext) {
```

src/HTTPClient.ts
```diff
@@ -33,6 +33,37 @@ export class HTTPClient {
...
+		return middlewareArray;
+	}
+
+	/**
+	 * @public
+	 * To set the middleware chain
+	 * @param {Middleware[]} middlewareArray - The array containing the middlewares
+	 */
+	public setMiddlewareArray(middlewareArray: Middleware[]) {
+		for (let num = 0; num < middlewareArray.length - 1; num += 1) {
```
baywet, line 61:
```suggestion
		middlewareArray.forEach((item, num) => {
```

src/HTTPClient.ts
```diff
@@ -33,6 +33,37 @@ export class HTTPClient {
...
+			middlewareArray[num].setNext(middlewareArray[num + 1]);
```
baywet, line 62:
```suggestion
			item.setNext(middlewareArray[num + 1]);
```

src/HTTPClient.ts
```diff
@@ -33,6 +33,37 @@ export class HTTPClient {
...
+		}
```
baywet, line 63:
```suggestion
		});
```

baywet (changes requested):

src/Version.ts
```diff
@@ -12,4 +12,4 @@
  * @module Version
  */
 
-export const PACKAGE_VERSION = "2.0.0";
+export const PACKAGE_VERSION = "2.1.0";
```
MIchaelMainer, line 15: I want to verify that we are working with the same code as the 2.1.0-preview.1 release. I expected to see this change from `2.1.0-preview.1` to `2.1.0`. What approach did you use to make this change?

nikithauc: I branched of the tag [2.1.0-preview.1](https://github.com/microsoftgraph/msgraph-sdk-javascript/tree/2.1.0-Preview.1) and then changed the version in package.json which is used to set version and created the PR against master.

The preview releases are made from dev. Since dev contains changes aside from  2.1.0-preview.1 I had to create a separate branch.
change from 2.1.0-preview.1 to 2.1.0 can be seen when [comparing with dev ](https://github.com/microsoftgraph/msgraph-sdk-javascript/compare/dev...nikithauc/Release-2.1.0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L3). 

I will merge dev with master after this release and push update the version in dev to preview 2.2.0.

MIchaelMainer: I see my confusion. I wasn't looking at package.json, and the package.json wasn't matching the tag. This was supposed to be changed in the last release but wasn't.

src/HTTPClient.ts
nikithauc: I will need to test this change more. I will release this in the next preview.

src/GraphRequestUtil.ts
nikithauc: Done.

baywet:
@nikithauc can I assume this one is replaced by #333

nikithauc:
@baywet  #333  is not a replacement. There will another PR for release2.1.0

PR #333, opened 2020-10-01
Client init with middleware array

nikithauc:
The latest preview release 2.1.0-Preview.1 has a features allowing user to create a middleware chain by passing an array([Commit](https://github.com/microsoftgraph/msgraph-sdk-javascript/commit/834fcf3e6bba7433b80370c6ce1a2d17483ae96a)). These changes were later removed from dev([Reverting Commit](https://github.com/microsoftgraph/msgraph-sdk-javascript/commit/aa17391479985c8cbe4ab4eb0093672a4db79aa2)). 

After a discussion with @darrelmiller and @MIchaelMainer, we planned to allow the user to pass an array of middleware  handlers while initializing the client.  This PR contains these changes.

After this PR is merged, I will be releasing a 2.1.0-Preview.2.

### DIFF
--- a/spec/core/Client.ts
+++ b/spec/core/Client.ts
@@ -8,7 +8,7 @@
 import { assert } from "chai";
 import "isomorphic-fetch";
 
-import { CustomAuthenticationProvider } from "../../src";
+import { CustomAuthenticationProvider, TelemetryHandler } from "../../src";
 import { Client } from "../../src/Client";
 import { AuthProvider } from "../../src/IAuthProvider";
 import { ClientOptions } from "../../src/IClientOptions";
@@ -75,6 +75,27 @@ describe("Client.ts", () => {
 			const responseBody = "Test response body";
 			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
 			const middlewareArray = [authHandler, new ChaosHandler(options)];
+			const client = Client.initWithMiddleware({ middleware: middlewareArray });
+
+			const response = await client.api("me").get();
+			assert.equal(response, responseBody);
+		});
+
+		it("Init middleware using a chained middleware array", async () => {
+			const provider: AuthProvider = (done) => {
+				done(null, "dummy_token");
+			};
+			const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
+
+			const responseBody = "Test response body";
+			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing chained middleware array", 0, responseBody);
+			const chaosHandler = new ChaosHandler(options);
+			const telemetryHandler = new TelemetryHandler();
+
+			authHandler.setNext(telemetryHandler);
+			telemetryHandler.setNext(chaosHandler);
+
+			const middlewareArray = [authHandler];
 			const client = Client.initWithMiddleware({ middleware: middlewareArray });
 
 			const response = await client.api("me").get();
--- a/spec/core/Client.ts
+++ b/spec/core/Client.ts
@@ -68,21 +68,17 @@ describe("Client.ts", () => {
 		});
 
 		it("Init middleware using a middleware array", async () => {
-			try {
-				const provider: AuthProvider = (done) => {
-					done(null, "dummy_token");
-				};
-				const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
-				const responseBody = "Test response body";
-				const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
-				const middlewareArray = [authHandler, new ChaosHandler(options)];
-				const client = Client.initWithMiddleware({ middleware: middlewareArray });
+			const provider: AuthProvider = (done) => {
+				done(null, "dummy_token");
+			};
+			const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
+			const responseBody = "Test response body";
+			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
+			const middlewareArray = [authHandler, new ChaosHandler(options)];
+			const client = Client.initWithMiddleware({ middleware: middlewareArray });
 
-				const response = await client.api("me").get();
-				assert.equal(response, responseBody);
-			} catch (error) {
-				assert.equal(error.name, "InvalidMiddlewareChain");
-			}
+			const response = await client.api("me").get();
+			assert.equal(response, responseBody);
 		});
 	});
 
--- a/spec/core/Client.ts
+++ b/spec/core/Client.ts
@@ -8,10 +8,15 @@
 import { assert } from "chai";
 import "isomorphic-fetch";
 
+import { CustomAuthenticationProvider } from "../../src";
 import { Client } from "../../src/Client";
 import { AuthProvider } from "../../src/IAuthProvider";
 import { ClientOptions } from "../../src/IClientOptions";
 import { Options } from "../../src/IOptions";
+import { AuthenticationHandler } from "../../src/middleware/AuthenticationHandler";
+import { ChaosHandler } from "../../src/middleware/ChaosHandler";
+import { ChaosHandlerOptions } from "../../src/middleware/options/ChaosHandlerOptions";
+import { ChaosStrategy } from "../../src/middleware/options/ChaosStrategy";
 import { DummyAuthenticationProvider } from "../DummyAuthenticationProvider";
 import { DummyHTTPMessageHandler } from "../DummyHTTPMessageHandler";
 
@@ -57,6 +62,24 @@ describe("Client.ts", () => {
 				const options: ClientOptions = {};
 				Client.initWithMiddleware(options);
 				throw new Error("Something wrong with the client initialization check");
+			} catch (error) {
+				assert.equal(error.name, "InvalidMiddlewareChain");
+			}
+		});
+
+		it("Init middleware using a middleware array", async () => {
+			try {
+				const provider: AuthProvider = (done) => {
+					done(null, "dummy_token");
+				};
+				const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
+				const responseBody = "Test response body";
+				const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
+				const middlewareArray = [authHandler, new ChaosHandler(options)];
+				const client = Client.initWithMiddleware({ middleware: middlewareArray });
+
+				const response = await client.api("me").get();
+				assert.equal(response, responseBody);
 			} catch (error) {
 				assert.equal(error.name, "InvalidMiddlewareChain");
 			}
--- a/spec/core/Client.ts
+++ b/spec/core/Client.ts
@@ -73,7 +73,7 @@ describe("Client.ts", () => {
 			};
 			const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
 			const responseBody = "Test response body";
-			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
+			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, "Testing middleware array", 200, 0, responseBody);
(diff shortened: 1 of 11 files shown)


src/tasks/OneDriveLargeFileUploadTask.ts
```diff
@@ -61,7 +61,10 @@ export class OneDriveLargeFileUploadTask extends LargeFileUploadTask {
 		}
 		// we choose to encode each component of the file path separately because when encoding full URI
 		// with encodeURI, special characters like # or % in the file name doesn't get encoded as desired
-		return `/me/drive/root:${path.split('/').map(p => encodeURIComponent(p)).join('/')}${encodeURIComponent(fileName)}:/createUploadSession`;
+		return `/me/drive/root:${path
```
nikithauc, line 64: This change is caused by the formatter.

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
+
+	/**
+	 * @private
+	 * Verifies if the middleware passed is an array or not before setting this.middleware property
+	 * @param {any} middleware - The middleware passed
+	 * @returns Nothing
+	 */
+
+	private setMiddleware(middleware: any): void {
+		if (Object.prototype.toString.call(middleware) === "[object Array]") {
```
MIchaelMainer, line 44: I think you can do `if Array.isArray(middleware)`

src/middleware/MiddlewareFactory.ts
```diff
@@ -0,0 +1,62 @@
...
+import { RetryHandler } from "./RetryHandler";
+import { TelemetryHandler } from "./TelemetryHandler";
+
+/**
+ * @private
+ * To check whether the environment is node or not
+ * @returns A boolean representing the environment is node or not
+ */
+const isNodeEnvironment = (): boolean => {
+	return new Function("try {return this === global;}catch(e){return false;}")(); // tslint:disable-line: function-constructor
```
MIchaelMainer, line 29: I'm curious as I don't know the answer. Why does this work better than checking if the node process is available or if there is a windows object?

src/middleware/MiddlewareFactory.ts
```diff
@@ -0,0 +1,62 @@
...
+	public static getDefaultMiddlewareChain(authProvider: AuthenticationProvider): Middleware[] {
+		const middleware: Middleware[] = new Array();
+		const authenticationHandler = new AuthenticationHandler(authProvider);
+		const retryHandler = new RetryHandler(new RetryHandlerOptions());
+		const telemetryHandler = new TelemetryHandler();
+		const httpMessageHandler = new HTTPMessageHandler();
+
+		middleware.push(authenticationHandler);
+		middleware.push(retryHandler);
+		if (isNodeEnvironment()) {
```
MIchaelMainer, line 53: Why do we not follow redirects in the browser? What ever the reason is, we'll need to do something similar for the .NET client in case of Blazor.

src/middleware/MiddlewareFactory.ts
nikithauc: Yes. I think that should work

src/HTTPClient.ts
nikithauc: I googled that `Array.isArray()` may not work in all environment and that is why used the condition above for array verification.

src/middleware/MiddlewareFactory.ts
nikithauc: This is what I learnt from @finsharp 
" Turns out that Browsers will automatically follow redirects in XHR requests, no matter the conditions. Browsers ALWAYS (as odd as it is) follow redirect, making it unnecessary to have the redirect handler registered"

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
+	}
+
+	/**
+	 * @private
+	 * Verifies if the middleware passed is an array or not before setting this.middleware property
+	 * @param {any} middleware - The middleware passed
+	 * @returns Nothing
+	 */
+
+	private setMiddleware(middleware: any): void {
```
baywet, line 43:
```suggestion
	private setMiddleware(...middleware: Middleware[]): void {
```

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
-	public constructor(middleware: Middleware) {
-		this.middleware = middleware;
+	public constructor(middleware: Middleware | Middleware[]) {
+		this.setMiddleware(middleware);
+	}
+
+	/**
+	 * @private
+	 * Verifies if the middleware passed is an array or not before setting this.middleware property
+	 * @param {any} middleware - The middleware passed
```
baywet, line 39:
```suggestion
	 * @param {Middleware} middleware - The middleware passed (variable)
```

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
+	 * @returns Nothing
+	 */
+
+	private setMiddleware(middleware: any): void {
+		if (Object.prototype.toString.call(middleware) === "[object Array]") {
+			this.parseMiddleWareArray(middleware);
+		} else {
+			this.middleware = middleware;
```
baywet, line 47:
```suggestion
			this.middleware = middleware[0];
```

src/HTTPClient.ts
baywet, line 44:
```suggestion
		if (middleware.length > 1) {
```

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
+	/**
+	 * @private
+	 * Processes the middleware array to construct the chain
+	 * and sets this.middleware property to the first middlware handler of the array
+	 * @param {Middleware[]} middlewareArray - The array of middleware handlers
+	 * @returns Nothing
+	 */
+	private parseMiddleWareArray(middlewareArray: Middleware[]) {
+		middlewareArray.forEach((element, index) => {
+			if (index < middlewareArray.length - 1) {
```
baywet, line 60: Thanks for integrating my last comment!

src/middleware/MiddlewareFactory.ts
```diff
@@ -0,0 +1,62 @@
...
+export class MiddlewareFactory {
+	/**
+	 * @public
+	 * @static
+	 * Returns the default middleware chain an array with the  middleware handlers
+	 * @param {AuthenticationProvider} authProvider - The authentication provider instance
+	 * @returns an array of the middleware handlers of the default middleware chain
+	 */
+	public static getDefaultMiddlewareChain(authProvider: AuthenticationProvider): Middleware[] {
+		const middleware: Middleware[] = new Array();
```
baywet, line 45:
```suggestion
		const middleware: Middleware[] = [];
```

baywet (changes requested):

src/HTTPClient.ts
nikithauc: > The middleware is the else condition is not an array.

baywet: @nikithauc this suggestion was to consider with the other ones, if you switch to var args it effectively becomes an array and simplifies your code

baywet: and provides a strongly type api surface that supports both scenarios (one or multiple middlewares)

src/HTTPClient.ts
nikithauc: > I feel the condition `Object.prototype.toString.call(middleware) === "[object Array]"`  indicates more clearly the purpose.

MIchaelMainer: Well I see now that Array.isArray is undefined in my browser.

baywet: that's a debatable opinion.
I think we should focus more on trying to keep as much as we can a strongly type API surface at every level to get compile time validation and more. `Middleware: any` is problematic here as it "looses" some of the compile time validation (both for the user of the method and for the implementation).
On top of that varargs are common practice in javascript/typescript allowing you to provide a clean, strongly typed API surface which serves multiple scenarios (single or multiple arguments)

baywet: See my other suggestion which avoids testing it this way all together.
But overall we should either:
- Document a minimum required version of ECMAScript and encourage people to polyfill (like we're doing we Promise)
- Setup a pipeline to transpile down to different ES versions 

I don't think using "older ES" in the source code is going to make the maintenance any easier for us/get us good perf on modern engines.

src/HTTPClient.ts
baywet: @MIchaelMainer it will be if you switch to varargs :)

MIchaelMainer: Ugh, my comment came after you two had the conversation. Wish GH would update the conversation without refresh

src/HTTPClient.ts
nikithauc: Using a spread operator requires more changes which I would like to take care of in a follow up PR.

nikithauc: @baywet  It is a good point to set consider multiple ES versions during build/test phase. 
Regarding support for older ES,  I will look into that.

src/HTTPClient.ts
nikithauc: This change could effect public interfaces. Will need to look into this separately to apply this with minimal changes to existing code.

baywet: as reviewed on the call, not necessarily, it could but if we want to avoid any public API change we could move up the type test.

spec/core/Client.ts
```diff
@@ -61,6 +66,20 @@ describe("Client.ts", () => {
...
+
+		it("Init middleware using a middleware array", async () => {
+			const provider: AuthProvider = (done) => {
+				done(null, "dummy_token");
+			};
+			const authHandler = new AuthenticationHandler(new CustomAuthenticationProvider(provider));
+			const responseBody = "Test response body";
+			const options = new ChaosHandlerOptions(ChaosStrategy.MANUAL, 200, "Testing middleware array", 0, responseBody);
+			const middlewareArray = [authHandler, new ChaosHandler(options)];
+			const client = Client.initWithMiddleware({ middleware: middlewareArray });
```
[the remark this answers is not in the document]
nikithauc: Added.

src/HTTPClient.ts
nikithauc: The type test is in private setMiddleware function so that the signature of public constructor of Client.ts is not changed.

src/Client.ts
```diff
@@ -93,7 +93,7 @@ export class Client {
 		} else if (clientOptions.authProvider !== undefined) {
 			httpClient = HTTPClientFactory.createWithAuthenticationProvider(clientOptions.authProvider);
 		} else if (clientOptions.middleware !== undefined) {
-			httpClient = new HTTPClient(clientOptions.middleware);
+			httpClient = new HTTPClient(...[].concat(clientOptions.middleware));
```
nikithauc, line 96: @ddyett  Thank you for your help with the code for integrating the spread operator!

src/HTTPClient.ts
nikithauc: Applied changes using the spread operator.

src/Client.ts
baywet: @nikithauc @ddyett great workaround! Thanks for taking the time to integrate the suggestions in the code base!

baywet (approved):

src/HTTPClient.ts
```diff
@@ -27,10 +27,41 @@ export class HTTPClient {
...
+	 * Processes the middleware parameter passed to set this.middleware property
+	 * @param {...Middleware} middleware - The middleware passed
+	 * @returns Nothing
+	 */
+
+	private setMiddleware(...middleware: Middleware[]): void {
+		if (middleware.length > 1) {
+			this.parseMiddleWareArray(middleware);
+		} else {
+			this.middleware = middleware[0];
```
zengin, line 47: This line will throw if the array is empty.

nikithauc: I did consider the case if the middleware is null or empty went with the assumption that if the user is not inputting a valid middleware sequence we shouldn't be letting the user continue and throw an error.

zengin: I checked with Typescript playground, if no parameter is provided to a varargs param, the indexing will resolve into undefined (must be a very good reason why they did that).

So it will be silent error, which will surface at another place. I think we should be more explicit here with this edge case.

Side note: even if it was throwing at this line, an explicit error message would have been better instead of throwing in the first line that tries to index into the array.

nikithauc: I added the check on the middleware parameter passed.

zengin (approved):

PR #334, opened 2020-10-01
Rearranging the parameters of  ChaosHandlerOptions constructor

nikithauc:
ChaosHandler feature is yet to be released, this change would not be  a breaking one.


MIchaelMainer (approved):
LGTM. Thank you for the parameter list order!

nikithauc:
> LGTM. Thank you for the parameter list order!

Thank you for pointing the problem with parameter list  in the first place!

PR #335, opened 2020-10-02
Make GraphError real Error

### LINKED ISSUE #248: GraphError should extend Error
Feature Request
Problem
Using this library in NodeJS Backend, most logger detect Error objects and parse them accordingly.
In addition, GraphError not being extended from Error doesn't have a stack trace.
GraphError is not parse correctly by our logger (pinoJS) and we don't have any stack trace.
Solution
Extend GraphError from base JS Error.
I believe it should be done here:

  
    
      msgraph-sdk-javascript/src/GraphError.ts
    
    
         Line 20
      in
      7529db8
    
    
    
    

        
          
           export class GraphError {
[AB#5832](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5832)

OlivierCuyp:
## Summary

Make GraphError a real JS Error with a stacktrace

## Motivation

Current GraphError is not a real Error and thus it misses the stacktrace.

## Test plan

## Closing issues

Fixes #248

## Types of changes

## Checklist

-   [x] I have updated the documentation accordingly.

### DIFF
--- a/spec/core/GraphErrorHandler.ts
+++ b/spec/core/GraphErrorHandler.ts
@@ -97,9 +97,9 @@ describe("GraphErrorHandler.ts", () => {
 
 		it("Should construct some default error", async () => {
 			const gError = await GraphErrorHandler.getError();
+			assert.equal(gError.message, "");
 			assert.equal(gError.statusCode, -1);
 			assert.equal(gError.code, null);
-			assert.equal(gError.message, null);
 			assert.equal(gError.body, null);
 			assert.equal(gError.requestId, null);
 		});
--- a/src/GraphError.ts
+++ b/src/GraphError.ts
@@ -17,7 +17,7 @@
  * Some fields are renamed ie, "request-id" => requestId so you can use dot notation
  */
 
-export class GraphError {
+export class GraphError extends Error {
 	/**
 	 * @public
 	 * A member holding status code of the error
@@ -29,12 +29,6 @@ export class GraphError {
 	 * A member holding code i.e name of the error
 	 */
 	public code: string | null;
-
-	/**
-	 * @public
-	 * A member holding error message
-	 */
-	public message: string | null;
 
 	/**
 	 * @public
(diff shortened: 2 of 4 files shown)


spec/core/GraphErrorHandler.ts
```diff
@@ -97,9 +97,9 @@ describe("GraphErrorHandler.ts", () => {
 
 		it("Should construct some default error", async () => {
 			const gError = await GraphErrorHandler.getError();
+			assert.equal(gError.message, "");
 			assert.equal(gError.statusCode, -1);
 			assert.equal(gError.code, null);
-			assert.equal(gError.message, null);
 			assert.equal(gError.body, null);
 			assert.equal(gError.requestId, null);
 		});
```
nikithauc, line 105: Can you please add unit tests verifying if stack is set or defined as expected as per the new line added in this PR -> 
`this.stack = baseError ? baseError.stack : this.stack;` ?

baywet (approved):

spec/core/GraphErrorHandler.ts
ddyett: probably need an issue to track adding a stack assertion for each of the cases here.

ddyett:
thanks for the contribution @OlivierCuyp

MIchaelMainer (approved):

OlivierCuyp:
@ddyett with pleasure

jviney:
I think this change has broken `instanceof` checks in Typescript. Previously a check like `err instanceof GraphError` would work as intended, now it always returns false.

See here for details on the Typescript issue:

https://github.com/microsoft/TypeScript/issues/13965
https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work

The issue can be worked around by editing the prototype as described in the TS docs above. This may not be considered a problem because it is standard TS behavior, but it is a breaking change from the previous version.

nikithauc:
@jviney [Version 2.1.1](https://www.npmjs.com/package/@microsoft/microsoft-graph-client/v/2.1.1) addresses the breaking change caused by the GraphError extending Error.

Thank you for bringing this issue to our notice! I appreciate it :)

PR #336, opened 2020-10-05
Add telemetry only for requests to graph.microsoft.com

### LINKED ISSUE #265: LargeFileUploadTask cors error
Description
Same issue as described at #220. I commented there as well, but the issue has been closed.
I'm having the exact same issue on the latest version 2.0.0. Only seems to happen for personal accounts trying to use LargeFileUploadTask for upload to onedrive. LargeFileUploadTask works fine when used with a business account uploading both to a sharepoint document library or onedrive.
Other requests like GET /drives/<drive-id>/items/children work perfectly.
Console Errors:
Access to fetch at 'https://api.onedrive.com/rup/24b5d554d8c0d12a/<longurlpartremoved>' from origin 'http://localhost:8081' has been blocked by CORS policy: Request header field sdkversion is not allowed by Access-Control-Allow-Headers in preflight response.
Steps to Reproduce
Using LargeFileUploadTask as described here
Expected behavior: Start file upload
Actual behavior: Fails with a Failed to fetch error message
Usage Information
SDK Version - 2.0.0

 Browser (Check, if using Browser version of SDK)

Browser Name - Chrome

Version - 79.0.3945.130 (64bit)
[AB#6002](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6002)

nikithauc:
Fixes #265

### DIFF
--- a/src/GraphRequestUtil.ts
+++ b/src/GraphRequestUtil.ts
@@ -15,6 +15,11 @@
 export const oDataQueryNames = ["$select", "$expand", "$orderby", "$filter", "$top", "$skip", "$skipToken", "$count"];
 
 /**
+ * To hold list of the service root endpoints for Microsoft Graph and Graph Explorer for each national cloud.
+ */
+export const graphURLs = ["https://graph.microsoft.com", "https://graph.microsoft.us", "https://dod-graph.microsoft.us", "https://graph.microsoft.de", "https://microsoftgraph.chinacloudapi.cn"];
+
+/**
  * To construct the URL by appending the segments with "/"
  * @param {string[]} urlSegments - The array of strings
  * @returns The constructed URL string
--- a/src/GraphRequestUtil.ts
+++ b/src/GraphRequestUtil.ts
@@ -17,7 +17,15 @@ export const oDataQueryNames = ["$select", "$expand", "$orderby", "$filter", "$t
 /**
  * To hold list of the service root endpoints for Microsoft Graph and Graph Explorer for each national cloud.
  */
-export const graphURLs = ["https://graph.microsoft.com", "https://graph.microsoft.us", "https://dod-graph.microsoft.us", "https://graph.microsoft.de", "https://microsoftgraph.chinacloudapi.cn"];
+export const graphURLs = new Set<string>();
+
+// using an IIFE to populate the set object with the graph host names as Set(iterable:Object) is not supported in Internet Explorer
+(() => {
+	const urls = ["graph.microsoft.com", "graph.microsoft.us", "dod-graph.microsoft.us", "graph.microsoft.de", "microsoftgraph.chinacloudapi.cn"];
+	urls.forEach((url) => {
+		graphURLs.add(url);
+	});
+})();
 
 /**
  * To construct the URL by appending the segments with "/"
--- a/src/middleware/TelemetryHandler.ts
+++ b/src/middleware/TelemetryHandler.ts
@@ -9,7 +9,7 @@
  * @module TelemetryHandler
  */
 
-import { GRAPH_BASE_URL } from "../Constants";
+import { graphURLs } from "../GraphRequestUtil";
 import { Context } from "../IContext";
 import { PACKAGE_VERSION } from "../Version";
 
@@ -65,9 +65,26 @@ export class TelemetryHandler implements Middleware {
 	 * @param {Context} context - The context object of the request
 	 * @returns A Promise that resolves to nothing
 	 */
+
+	/**
+	 * @private
+	 * Checks if the request url points to the Graph service endpoints
+	 * @param {string} url - The request url string
+	 * @returns true if request url is a Graph url
+	 */
+	private isGraphURL(url: string): boolean {
+		let isGraph = false;
+		graphURLs.forEach((element) => {
+			if (url.indexOf(element) !== -1) {
+				isGraph = true;
+			}
+		});
+		return isGraph;
+	}
+
 	public async execute(context: Context): Promise<void> {
 		try {
-			if (typeof context.request === "string" && context.request.indexOf(GRAPH_BASE_URL) !== -1) {
+			if (typeof context.request === "string" && this.isGraphURL(context.request)) {
 				let clientRequestId: string = getRequestHeader(context.request, context.options, TelemetryHandler.CLIENT_REQUEST_ID_HEADER);
 				if (clientRequestId === null) {
 					clientRequestId = generateUUID();
--- a/src/middleware/TelemetryHandler.ts
+++ b/src/middleware/TelemetryHandler.ts
@@ -9,6 +9,7 @@
  * @module TelemetryHandler
  */
 
+import { GRAPH_BASE_URL } from "../Constants";
 import { Context } from "../IContext";
 import { PACKAGE_VERSION } from "../Version";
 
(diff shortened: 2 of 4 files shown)


src/middleware/TelemetryHandler.ts
```diff
@@ -66,21 +67,25 @@ export class TelemetryHandler implements Middleware {
...
-			}
-			let sdkVersionValue: string = `${TelemetryHandler.PRODUCT_NAME}/${PACKAGE_VERSION}`;
-			let options: TelemetryHandlerOptions;
-			if (context.middlewareControl instanceof MiddlewareControl) {
-				options = context.middlewareControl.getMiddlewareOptions(TelemetryHandlerOptions) as TelemetryHandlerOptions;
-			}
-			if (typeof options !== "undefined") {
-				const featureUsage: string = options.getFeatureUsage();
-				sdkVersionValue += ` (${TelemetryHandler.FEATURE_USAGE_STRING}=${featureUsage})`;
+			if (typeof context.request === "string" && context.request.indexOf(GRAPH_BASE_URL) !== -1) {
```
baywet, line 70: this is not going to work for national deployments

baywet, line 70:
```suggestion
			if (context.request instanceof string && context.request.indexOf(GRAPH_BASE_URL) !== -1) {
```

baywet: us gov, chinese cloud, germany, etc...
https://docs.microsoft.com/en-us/graph/deployments

nikithauc: Will checking if the url has the keyword 'graph' cover all the cases? 
If yes, could there be a case where the url contains 'graph' but the telemetry shouldn't be passed?

baywet: I did something similar for unit testing for the mock test auth provider for large files upload to avoid attaching the JWT on non-graph requests. The only issue I see with checking just for graph is that teams has a service called teamsgraph.something.com (don't remember the full hostname) and some services have graph in the API path.
So I think it might be safer to check for "graph" only in the hostname as opposed to the full URL. But it might still cause us issues down the road if AGS enables more pass-through than today...

An alternative could be to put all the "known graph hostnames" in a array and compare if we have a match with anything in the array. A bit more compute intensive and it'll require us to update the array as new national deployments come online. But definitively safer.

nikithauc: https://stackoverflow.com/questions/899574/what-is-the-difference-between-typeof-and-instanceof-and-when-should-one-be-used
[Example from TypeScript handbook](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#control-flow-based-type-analysis)

src/middleware/TelemetryHandler.ts
```diff
@@ -64,23 +65,44 @@ export class TelemetryHandler implements Middleware {
...
 	 * @returns A Promise that resolves to nothing
 	 */
+
+	/**
+	 * @private
+	 * Checks if the request url points to the Graph service endpoints
+	 * @param {string} url - The request url string
+	 * @returns true if request url is a Graph url
+	 */
+	private isGraphURL(url: string): boolean {
```
zengin, line 75: 1. construct a `Set()` object for `graphURLs` where the values are only the host names.
2. construct a typed `URL` object for `url`.
3. Have this function return the result of `graphURLs.has(urlObject.hostname)`

Benefits:
1. Efficiency: Lookup is amortized O(1).
2. Correctness: We probably want to capture telemetry for urls like: `https://graph.microsoft.com:443` and current implementation will not accept URLs with port numbers.

zengin: also consider inlining the whole function as it will be a simple call to `graphURLs.has(new URL(context.request).hostname)`

zengin (changes requested):

src/middleware/TelemetryHandler.ts
nikithauc: Thank you for suggesting an optimal approach!

I have used the approach you suggested with some modifications. Implemented a set object with the host names. I haven't used the URL object as  some browsers do not support of the URL object and I am extracting the host name as it is done in [GraphRequest.ts](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/66a41814cace93cf1da8d5ead69b8df86c1881b0/src/GraphRequest.ts#L135) during the url parsing process.

src/GraphRequestUtil.ts
```diff
@@ -17,7 +17,15 @@ export const oDataQueryNames = ["$select", "$expand", "$orderby", "$filter", "$t
 /**
  * To hold list of the service root endpoints for Microsoft Graph and Graph Explorer for each national cloud.
  */
-export const graphURLs = ["https://graph.microsoft.com", "https://graph.microsoft.us", "https://dod-graph.microsoft.us", "https://graph.microsoft.de", "https://microsoftgraph.chinacloudapi.cn"];
+export const graphURLs = new Set<string>();
+
+// using an IIFE to populate the set object with the graph host names as Set(iterable:Object) is not supported in Internet Explorer
+(() => {
+	const urls = ["graph.microsoft.com", "graph.microsoft.us", "dod-graph.microsoft.us", "graph.microsoft.de", "microsoftgraph.chinacloudapi.cn"];
```
baywet, line 24: you could initialize the set directly from the array, see the edit in the response.
https://stackoverflow.com/a/30902786/3808675

src/middleware/TelemetryHandler.ts
```diff
@@ -73,13 +73,23 @@ export class TelemetryHandler implements Middleware {
...
+
+		if (url.indexOf("https://") !== -1) {
+			url = url.replace("https://", "");
+
+			// Find where the host ends
+			const endOfHostStrPos = url.indexOf("/");
+			if (endOfHostStrPos !== -1) {
+				// Parse out the host
+				const hostName = url.substring(0, endOfHostStrPos);
+				return graphURLs.has(hostName);
```
baywet, line 87: what happens if the hostname part of the URL has capital letters in it? (like GRAPH.microsoft.com)

baywet (changes requested):

src/GraphRequestUtil.ts
nikithauc: I used an IIFE with set.add() instead as IE does not support Set(array). IE creates an empty set when I passed in new Set(object/array).

src/middleware/TelemetryHandler.ts
nikithauc: Made the comparison to be case insensitive.

src/GraphRequestUtil.ts
baywet: interesting, and I'm guessing this is an edge case the transpilation is not catching?

baywet: additionally this is something that people (or we) can polyfill using core-js, see the first example. https://www.npmjs.com/package/core-js

src/middleware/TelemetryHandler.ts
```diff
@@ -64,23 +65,55 @@ export class TelemetryHandler implements Middleware {
...
+		url = url.toLowerCase();
+
+		if (url.indexOf("https://") !== -1) {
+			url = url.replace("https://", "");
+
+			// Find where the host ends
+			const endOfHostStrPos = url.indexOf("/");
+			if (endOfHostStrPos !== -1) {
+				// Parse out the host
+				const hostName = url.substring(0, endOfHostStrPos);
```
zengin, line 87: this doesn't cover the scenario of `https://graph.microsoft.com:443/v1.0/me`

nikithauc:
Will create another pull request dealing with custom headers.

Issue #342, opened 2020-10-12
Post /subscriptions does not work


vemundeldegard:
# Bug Report

## Prerequisites

-   [x] Can you reproduce the problem?
-   [x] Are you running the latest version?
-   [x] Are you reporting to the correct repository?
-   [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

## Description
I am trying to create a new subscription by using this client. It does not work and throws an error with no information. Error code -1. 

Creating it by sending a manual request to the Microsoft Graph API works fine. 

**Console Errors:** [Is there any console error]
```
communication-ms    |     Error
communication-ms    |     at new GraphError (/app/node_modules/@microsoft/microsoft-graph-client/lib/src/GraphError.js:31:28)
communication-ms    |     at Function.<anonymous> (/app/node_modules/@microsoft/microsoft-graph-client/lib/src/GraphErrorHandler.js:98:30)
communication-ms    |     at step (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:141:27)
communication-ms    |     at Object.next (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:122:57)
communication-ms    |     at /app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:115:75
communication-ms    |     at new Promise (<anonymous>)
communication-ms    |     at Object.__awaiter (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:111:16)
communication-ms    |     at Function.GraphErrorHandler.getError (/app/node_modules/@microsoft/microsoft-graph-client/lib/src/GraphErrorHandler.js:88:24)
communication-ms    |     at GraphRequest.<anonymous> (/app/node_modules/@microsoft/microsoft-graph-client/lib/src/GraphRequest.js:305:84)
communication-ms    |     at step (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:141:27)
communication-ms    |     at Object.throw (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:122:57)
communication-ms    |     at rejected (/app/node_modules/@microsoft/microsoft-graph-client/node_modules/tslib/tslib.js:113:69)
communication-ms    |     at processTicksAndRejections (internal/process/task_queues.js:93:5) {
[...]

PR #353, opened 2020-11-10
Empty error reported in the SDK

nikithauc:
#342 

Problem - 
1. The issue reported states that the response was empty with a -1 status code. I was unable to reproduce this issue for "/subscriptions". 

However, I was able to get an empty response when I tried the following example to set the client-
```
const client = MicrosoftGraph.Client.init({
	defaultVersion: "v1.0",
	debugLogging: true,
	authProvider: (done) => {
		done(null, "");
	},
});
then try client.api("/me").get().
```

`export type AuthProviderCallback = (error: any, accessToken: string | null) => void;`  If the implementation 
 of AuthProviderCallback created by the consumer returns an empty string for an access token and the error is set null, the response by the Graph SDK is an empty one with status code as -1 which could be confusing.
2.  The GraphErrorHandler takes in error parameter of type 'any' but didn't handle, say error of type 'string', correctly.

Solution in this PR -
1. If the access token is empty and the error specified is undefined/null, create a GraphError stating that the AccessToken is needs to specified.
2. Also,  modified the GraphErrorHandler to handle `any` type error parameter that is passed appropriately .

### DIFF
--- a/src/CustomAuthenticationProvider.ts
+++ b/src/CustomAuthenticationProvider.ts
@@ -47,6 +47,9 @@ export class CustomAuthenticationProvider implements AuthenticationProvider {
 				if (accessToken) {
 					resolve(accessToken);
 				} else {
+					if (!error) {
+						reject("Access token cannot be undefined or empty");
+					}
 					reject(error);
 				}
 			});
--- a/src/CustomAuthenticationProvider.ts
+++ b/src/CustomAuthenticationProvider.ts
@@ -49,9 +49,12 @@ export class CustomAuthenticationProvider implements AuthenticationProvider {
 					resolve(accessToken);
 				} else {
 					if (!error) {
-						error = new GraphClientError("Access token cannot be undefined or empty.");
+						const invalidTokenMessage = "Access token is undefined or empty.\
+						Please provide a valid token.\
+						For more help - https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/docs/CustomAuthenticationProvider.md";
+						error = new GraphClientError(invalidTokenMessage);
 					}
-					const err = await GraphClientError.getError(error);
+					const err = await GraphClientError.setGraphClientError(error);
 					reject(err);
 				}
 			});
--- a/src/CustomAuthenticationProvider.ts
+++ b/src/CustomAuthenticationProvider.ts
@@ -9,7 +9,7 @@
  * @module CustomAuthenticationProvider
  */
 
-import { GraphError } from "./browser";
+import { GraphClientError } from "./GraphClientError";
 import { AuthenticationProvider } from "./IAuthenticationProvider";
 import { AuthProvider } from "./IAuthProvider";
 
@@ -44,14 +44,15 @@ export class CustomAuthenticationProvider implements AuthenticationProvider {
 	 */
 	public async getAccessToken(): Promise<any> {
 		return new Promise((resolve: (accessToken: string) => void, reject: (error: any) => void) => {
-			this.provider((error: any, accessToken: string | null) => {
+			this.provider(async (error: any, accessToken: string | null) => {
 				if (accessToken) {
 					resolve(accessToken);
 				} else {
 					if (!error) {
-						error = new GraphError(-1, "Access token cannot be undefined or empty.");
+						error = new GraphClientError("Access token cannot be undefined or empty.");
 					}
-					reject(error);
+					const err = await GraphClientError.getError(error);
+					reject(err);
 				}
 			});
 		});
--- a/src/CustomAuthenticationProvider.ts
+++ b/src/CustomAuthenticationProvider.ts
@@ -9,6 +9,7 @@
  * @module CustomAuthenticationProvider
  */
 
+import { GraphError } from "./browser";
 import { AuthenticationProvider } from "./IAuthenticationProvider";
 import { AuthProvider } from "./IAuthProvider";
 
@@ -48,7 +49,7 @@ export class CustomAuthenticationProvider implements AuthenticationProvider {
 					resolve(accessToken);
 				} else {
 					if (!error) {
-						reject("Access token cannot be undefined or empty");
+						error = new GraphError(-1, "Access token cannot be undefined or empty.");
 					}
 					reject(error);
 				}
--- a/src/GraphErrorHandler.ts
+++ b/src/GraphErrorHandler.ts
@@ -13,6 +13,20 @@ import { GraphError } from "./GraphError";
 import { GraphRequestCallback } from "./IGraphRequestCallback";
 
 /**
+ * @interface
+ * Signature for the json represent of the error response from the Graph API
+ * https://docs.microsoft.com/en-us/graph/errors
+ * @property {[key: string] : string | number} - The Key value pair
+ */
+interface GraphAPIErrorResponse {
+	error: {
+		code: string;
+		message: string;
+		innerError: any;
+	};
+}
+
+/**
  * @class
  * Class for GraphErrorHandler
  */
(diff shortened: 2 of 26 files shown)


src/GraphErrorHandler.ts
```diff
@@ -85,10 +91,10 @@ export class GraphErrorHandler {
 	 */
 	public static async getError(error: any = null, statusCode: number = -1, callback?: GraphRequestCallback): Promise<GraphError> {
 		let gError: GraphError;
-		if (error && error.error) {
-			gError = GraphErrorHandler.constructErrorFromResponse(error, statusCode);
-		} else if (typeof Error !== "undefined" && error instanceof Error) {
+		if (typeof Error !== "undefined" && error instanceof Error) {
```
baywet, line 94:
```suggestion
		if (Error !== undefined && error instanceof Error) {
```

baywet (changes requested):

src/CustomAuthenticationProvider.ts
```diff
@@ -47,6 +48,9 @@ export class CustomAuthenticationProvider implements AuthenticationProvider {
 				if (accessToken) {
 					resolve(accessToken);
 				} else {
+					if (!error) {
+						error = new GraphError(-1, "Access token cannot be undefined or empty.");
```
MIchaelMainer, line 52: We really should have a GraphClientError class since GraphError is modeled around OData errors. A GraphClientError object should provide actionable information since we typically know what caused an error. Ideally, we provide URL with an aka.ms link to information that the developer needs to resolve the issue.

If we know how the access token may have gotten in the undefined or empty state, we should provide guidance on how to fix this.

src/GraphErrorHandler.ts
```diff
@@ -58,6 +58,12 @@ export class GraphErrorHandler {
 	 *  }
 	 */
 	private static constructErrorFromResponse(error: any, statusCode: number): GraphError {
+		if (!error.error) {
+			// example - Consider error specified is a string when using type AuthProviderCallback = (error: any, accessToken: string | null) => void;
+			const graphError = new GraphError(statusCode, "Microsoft Graph JavaScript Client SDK came across an error while processing the request");
```
zengin, line 63: Where does the statusCode value come from? It looks undefined at this line.

src/GraphErrorHandler.ts
```diff
@@ -58,6 +58,12 @@ export class GraphErrorHandler {
 	 *  }
 	 */
 	private static constructErrorFromResponse(error: any, statusCode: number): GraphError {
+		if (!error.error) {
```
zengin, line 61: values of error being `undefined` or `null` are not handled.

src/GraphErrorHandler.ts
```diff
@@ -58,6 +58,12 @@ export class GraphErrorHandler {
...
+			// example - Consider error specified is a string when using type AuthProviderCallback = (error: any, accessToken: string | null) => void;
+			const graphError = new GraphError(statusCode, "Microsoft Graph JavaScript Client SDK came across an error while processing the request");
+			graphError.body = error;
+			return graphError;
+		}
 		error = error.error;
```
zengin, line 67: Can we call the function parameter outerError, and this one innerError or something more descriptive of what these objects look like? `error = error.error` is very confusing.

src/GraphErrorHandler.ts
zengin: Is `error` guaranteed to be a valid object?

src/GraphErrorHandler.ts
zengin: never mind, I missed the function parameter.

src/GraphErrorHandler.ts
nikithauc: Changed the parameter name passed to the function to graphError

src/CustomAuthenticationProvider.ts
nikithauc: Added a GraphClientError class

src/GraphErrorHandler.ts
nikithauc: This function is called if error.error exists.

zengin (approved):

src/GraphErrorHandler.ts
```diff
@@ -57,25 +57,17 @@ export class GraphErrorHandler {
...
 	 */
-	private static constructErrorFromResponse(error: any, statusCode: number): GraphError {
-		if (!error.error) {
-			// example - Consider error specified is a string when using type AuthProviderCallback = (error: any, accessToken: string | null) => void;
-			const graphError = new GraphError(statusCode, "Microsoft Graph JavaScript Client SDK came across an error while processing the request");
-			graphError.body = error;
-			return graphError;
-		}
-		error = error.error;
+	private static constructErrorFromResponse(graphError: any, statusCode: number): GraphError {
```
baywet, line 60: add a type to avoid any and get transpile time checks?

src/GraphErrorHandler.ts
```diff
@@ -84,19 +76,21 @@ export class GraphErrorHandler {
 	 * @static
 	 * @async
 	 * To get the GraphError object
+	 * Reference - https://docs.microsoft.com/en-us/graph/errors
 	 * @param {any} [error = null] - The error returned by graph service or some native error
 	 * @param {number} [statusCode = -1] - The status code of the response
 	 * @param {GraphRequestCallback} [callback] - The graph request callback function
 	 * @returns A promise that resolves to GraphError instance
 	 */
 	public static async getError(error: any = null, statusCode: number = -1, callback?: GraphRequestCallback): Promise<GraphError> {
```
baywet: type use

src/GraphRequest.ts
```diff
@@ -379,7 +379,11 @@ export class GraphRequest {
 			const response: any = await GraphResponseHandler.getResponse(rawResponse, this._responseType, callback);
 			return response;
 		} catch (error) {
+			if (error instanceof GraphClientError) {
+				throw error;
+			}
 			let statusCode: number;
+
 			if (typeof rawResponse !== "undefined") {
```
baywet, line 387:
```suggestion
			if (rawResponse) {
```

baywet (changes requested):

src/GraphErrorHandler.ts
nikithauc: Can you please elaborate on transpile time checks?

baywet: right now, because graphError is of any type, none of the things referenced under the graphError parameter are being checked by the compiler/transpiler.
Effectively if somebody renames error to errorObject or errorPayload in the source object that's being passed here, this method will break.
Not only it will break, but it won't say anything at compile time and break at runtime.
Using any also doesn't help consumers as they don't know what this method is expecting.

nikithauc: 1. This is a private method which is specifically designed around the Graph API Error response. I agree that the code should be loosely coupled. I will create an issue after finding  more on this.
2. The error passed to the GraphErrorHandler could be a Graph API Error or could also be any custom error (say if  using a custom middleware).  GraphErrorHandler has been taking into consideration both a Graph API Error and any custom error.

src/GraphErrorHandler.ts
nikithauc: Same reason as above.  We need to consider any custom error that consumers currently can be passing. Strict type use would be a breaking change.

zengin (dismissed):

src/GraphErrorHandler.ts
baywet: if we don't want to replace any by something else here for the reasons you have exposed, we should at least test the properties in use before using them to avoid errors at runtime.
Things like testing if graphError.error exists before assigning it (defensive programming) or using null propagation operators (graphError && graphError.error).
Another alternative would be to replace any by a prototype of what that method accesses i.e. something like that
`{error: {message: string, code: number}}...`

src/GraphErrorHandler.ts
```diff
@@ -41,7 +55,7 @@ export class GraphErrorHandler {
 	 * @static
 	 * @async
 	 * Populates the GraphError instance from the Error returned by graph service
-	 * @param {any} error - The error returned by graph service or some native error
+	 * @param {any} graphError - The error possibly returned by graph service or some native error
```
baywet, line 58:
```suggestion
	 * @param {GraphAPIErrorResponse} graphError - The error possibly returned by graph service or some native error
```

baywet (changes requested):

baywet (approved):

PR #351, opened 2020-11-10
Adding/Deleting authorization and telemetry headers

nikithauc:
Fixes #265 and #344 

Problem 1 - #265 reports a CORs error which was caused because the SDK added telemetry headers only for every request that goes out. The redirected non-Graph endpoint complained about the unexpected SDK_Version header and caused a CORS error.
Solution for Problem 1 - This PR includes logic to check in the `TelemetryHandler` if the URL is a Graph URL and if yes process to add the telemetry headers. If the URL is a non Graph URL and the telemetry headers are present, then the headers are deleted.

Problem 2 - #344 @peombwa Thank you for helping me reproduce this issue! 
On redirection, the `RedirectionHandler` is currently setting the AuthorizationHeader to undefined instead of completely dropping the header key. 
The existing code worked fine in a node environment but failed when using ExpressJS. 

Solution for Problem 2 - This PR includes change in the `RedirectionHandler` to delete Authorization header key if present.

PR #357, opened 2020-11-18
Checking if the request object url is a graph url

nikithauc:
in #351, I made changes to add/drop telemetry headers if the request URL is /isn't a Graph URL.

Added one more small change in the `TelemetryHandler` to read the request url from the context.request(which could be a string or a Request object) before checking if the url is a Graph URL or not.

### DIFF
--- a/src/middleware/TelemetryHandler.ts
+++ b/src/middleware/TelemetryHandler.ts
@@ -71,7 +71,7 @@ export class TelemetryHandler implements Middleware {
 				// Add telemetry only if the request url is a Graph URL.
 				// Errors are reported as in issue #265 if headers are present when redirecting to a non Graph URL
 				let clientRequestId: string = getRequestHeader(context.request, context.options, TelemetryHandler.CLIENT_REQUEST_ID_HEADER);
-				if (clientRequestId === null) {
+				if (!clientRequestId) {
 					clientRequestId = generateUUID();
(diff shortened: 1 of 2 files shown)


zengin (dismissed):

src/middleware/TelemetryHandler.ts
```diff
@@ -66,30 +66,29 @@ export class TelemetryHandler implements Middleware {
...
-				} else {
-					// Remove telemetry headers if present during redirection.
-					delete context.options.headers[TelemetryHandler.CLIENT_REQUEST_ID_HEADER];
-					delete context.options.headers[TelemetryHandler.SDK_VERSION_HEADER];
+			const url = typeof context.request === "string" ? context.request : context.request.url;
+			if (isGraphURL(url)) {
+				// Add telemetry only if the request url is a Graph URL.
+				// Errors are reported as in issue #265 if headers are present when redirecting to a non Graph URL
+				let clientRequestId: string = getRequestHeader(context.request, context.options, TelemetryHandler.CLIENT_REQUEST_ID_HEADER);
+				if (clientRequestId === null) {
```
baywet, line 74:
```suggestion
				if (!clientRequestId) {
```

baywet (changes requested):

baywet (approved):
Thanks for including the suggestion

zengin (approved):

Issue #230, opened 2019-09-22
ImplicitMSALAuthenticationProvider is not exported in index.js


darrelmiller:
We should add this file to the list of files we export.

Currently you need to export the file explicitly like this:

```js
import { ImplicitMSALAuthenticationProvider }  from "@microsoft/microsoft-graph-client/lib/src/ImplicitMSALAuthenticationProvider";

```

Ideally we should be able to do

```js
import * as MicrosoftGraph from "@microsoft/microsoft-graph-client";
```

pl4nty:
Worth noting MSALAuthenticationProviderOptions isn't exported either, even though the Authentication section of the readme uses `new MicrosoftGraph.MSALAuthenticationProviderOptions(graphScopes);`

sowvikr:
How the code is not reviewed and merged into the master?

JackAdams:
> Worth noting MSALAuthenticationProviderOptions isn't exported either, even though the Authentication section of the readme uses `new MicrosoftGraph.MSALAuthenticationProviderOptions(graphScopes);`

So how do you access `MSALAuthenticationProviderOptions`?  Is it this:

```
import { MSALAuthenticationProviderOptions }  from '@microsoft/microsoft-graph-client/lib/src/MSALAuthenticationProviderOptions'
```

PR #370, opened 2021-01-04
Unit tests restructuring

nikithauc:
This PR focusses to improve the existing test structure of the project. Changes are as follows -
1. Added testing tools -
		a. `Karma` - To test the unit test in browsers
		b. `@istanbuljs/nyc-config-typescript` - Evaluation of the test coverage
2. Restructured the test folders as follows 
		a. Renamed from folder spec/  to test/
		b. Divided unit test in folders - common, node and browser.
		c. Removed the config files from the test folder.
3. The JS SDK is transpiled into CommonJs and ESM modules. Added `esm` dependency to test run mocha tests on transpiled ESM modules.
4. Some tests were failing when I was using the latest version of isomorphic-fetch/ cross-fetch/via karma. Updated those  tests to align to the updates in the `fetch api`. Updated  the isomorphic-fetch dependency. 
5. Corrected some invalid tests.   
6. Added headers as a property in the ChaosHandlerOptions. I plan for some more the ChaosHandlerOptions design changes but not in this PR.

TODO - Check if the any changes are needed in the GraphResponseHandler to adhere to fetch api/http/ Graph Response standards. Will create a follow up PR if any changes are present

PR #371, opened 2021-01-04
@azure/identity-TokenCredential

nikithauc:
1. Introduces `TokenCredentialAuthenticationProvider` option so that authentication can be done using the `@azure/identity- tokenCredential` classes.
2. Added samples illustration the use of `TokenCredentialAuthenticationProvider` using `AuthenticationCodeFlow and ClientCredentialFlow`
3. Added and updated documents.
4. Introducing the authentication subfolder in src/   changing the structure of the source folders.
5. Added authProviderOptions/ folder which exports the msal and TokenCredential authentication provider options. This is done so that the we can use  import {TokenCredentialAuthenticationProvider} from`@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials`. This is a workaround for the limitations https://github.com/microsoft/TypeScript/issues/8305 ,
https://github.com/microsoft/TypeScript/issues/33079. 

The idea for using a submodule structure to export the authProviderOptions is because -  
1. We are not exporting `TokenCredentialAuthenticationProvider` and `ImplicitMsalAuthenticationProvider` in the `src/index.ts`. This makes `msal` and `@azure/identity` an optional dependency for the user of the sdk.  Marking them as option using [peerDependenciesMeta](https://pnpm.js.org/en/package_json#peerdependenciesmeta).
2. #230 - `ImplicitMsalAuthenticationProvider` was not exported in JS SDK version < 2.1 as using `msal ` was optional but exporting in the index file would make `msal ` a runtime dependency.
3. To import `TokenCredentialAuthenticationProvider` and `ImplicitMsalAuthenticationProvider`, the paths would be 

        const {TokenCredentialAuthenticationProvider} = require("@microsoft/microsoft-graph- 
        client/authProviders/azureTokenCredentials");
        
        const{ImplicitMSALAuthenticationProvider}  = require("@microsoft/microsoft-graph-client/authProviders/msal");
4. Added `tsconfig-sub-x.json` to build the project using TypeScript [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html)


nikithauc:
This PR depends on #370

README.md
```diff
@@ -55,65 +55,13 @@ Register your application to use Microsoft Graph API using one of the following
...
-};
-const graphScopes = ["user.read", "mail.send"]; // An array of graph scopes
-
-// Important Note: This library implements loginPopup and acquireTokenPopup flow, remember this while initializing the msal
-// Initialize the MSAL @see https://github.com/AzureAD/microsoft-authentication-library-for-js#1-instantiate-the-useragentapplication
-const msalApplication = new UserAgentApplication(msalConfig);
-const options = new MSALAuthenticationProviderOptions(graphScopes);
-const authProvider = new ImplicitMSALAuthenticationProvider(msalApplication, options);
-```
+    > **Important Note:** MSAL is supported only for frontend applications, for server-side authentication you have to implement your own AuthenticationProvider. Learn how you can create a [Custom Authentication Provider](./docs/CustomAuthenticationProvider.md).
```
baywet, line 64: Can't they use [msal node](https://www.npmjs.com/package/@azure/msal-node)?

.gitignore
```diff
@@ -14,10 +14,13 @@ npm-debug.log
 /lib/*
 !/lib/.npmignore
 
+/authProviders/*
```
baywet, line 17: I'm not sure why this path is excluded when it contains MD files?

docs/ImplicitMSALAuthenticationProvider.md
```diff
@@ -0,0 +1,54 @@
...
+const options = new MicrosoftGraph.MSALAuthenticationProviderOptions(graphScopes);
+const authProvider = new MicrosoftGraph.ImplicitMSALAuthenticationProvider(msalApplication, options);
+```
+
+#### Creating an instance of ImplicitMSALAuthenticationProvider in node environment
+
+Refer devDependencies in [package.json](./package.json) for the compatible msal version and update that version in below.
+
+```cmd
+npm install msal@<version>
```
baywet, line 31:
```suggestion
npm install msal@<version> --save
```

package.json
```diff
@@ -3,28 +3,38 @@
...
+		"src/",
+		"authProviders/"
 	],
 	"types": "./lib/src/index.d.ts",
 	"dependencies": {
 		"@babel/runtime": "^7.4.4",
 		"tslib": "^1.9.3"
 	},
 	"devDependencies": {
+		"@azure/identity": "^1.1.0",
```
baywet, line 27: shouldn't azure identity be in our deps and not dev deps since we have a hard dependency at runtime on it?

samples/tokenCredentialSamples/AuthenticationCodeFlow/package.json
```diff
@@ -0,0 +1,16 @@
...
+	"scripts": {
+		"start": "npm install && npx ts-node src/index.ts"
+	},
+	"dependencies": {
+		"@azure/identity": "^1.1.0",
+		"express": "^4.17.1",
+		"open": "^7.2.1",
+		"qs": "^6.9.4",
+		"ts-node": "^9.0.0",
+		"typescript": "^4.0.3"
```
baywet, line 14: shouldn't our sdk be in the list of deps, but with a relative path instead?

spec/core/GraphResponseHandler.ts
```diff
@@ -66,14 +66,15 @@ describe("GraphResponseHandler.ts", () => {
...
 	/* tslint:disable: no-string-literal */
-	describe("parseDocumentResponse", () => {
-		it("Should return the html string", async () => {
-			const response = new Response(htmlString, status200);
-			const dom = await GraphResponseHandler["parseDocumentResponse"](response, DocumentType.TEXT_HTML);
-			assert.isDefined(dom);
-			assert.equal(typeof dom, "string");
-		});
-	});
+	// Test Correction required
```
baywet, line 69: is it disabled by intention or was it forgotten?

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * @public
+	 * @constructor
+	 * Creates an instance of TokenCredentialAuthenticationProvider
+	 * @param {TokenCredential} tokenCredential - An instance of @azure/identity TokenCredential
+	 * @param {TokenCredentialAuthenticationProviderOptions} authenticationProviderOptions - An instance of TokenCredentialAuthenticationProviderOptions
+	 * @returns An instance of TokenCredentialAuthenticationProvider
+	 */
+	public constructor(tokenCredential: TokenCredential, authenticationProviderOptions: TokenCredentialAuthenticationProviderOptions) {
+		this.authenticationProviderOptions = authenticationProviderOptions;
```
baywet, line 47: defensive programming?

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * @public
+	 * @async
+	 * To get the access token
+	 * @param {TokenCredentialAuthenticationProviderOptions} authenticationProviderOptions - The authentication provider options object
+	 * @returns The promise that resolves to an access token
+	 */
+	public async getAccessToken(): Promise<string> {
+		let scopes: string[] = [];
```
baywet, line 59:
```suggestion
		let scopes: string[] = this.authenticationProviderOptions?.scopes ?? [];
```

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+		const error = new GraphClientError();
+		if (this.authenticationProviderOptions && this.authenticationProviderOptions.scopes) {
```
baywet, line 61:
```suggestion
```

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+			scopes = this.authenticationProviderOptions.scopes;
```
baywet, line 62:
```suggestion
```

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
```
baywet, line 63:
```suggestion
```

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+		if (scopes === undefined || scopes === null || scopes.length === 0) {
```
baywet, line 64:
```suggestion
		if (scopes.length === 0) {
```

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+			error.name = "Empty Scopes";
+			error.message = "Scopes cannot be empty, Please provide scopes";
+			throw error;
+		}
+		const response = await this.tokenCredential.getToken(scopes, this.authenticationProviderOptions.getTokenoptions);
```
baywet, line 69: this will throw if tokenCredential is null (see my comment on the constructor about defensive programming)

src/index.ts
```diff
@@ -40,7 +40,5 @@ export * from "./IClientOptions";
 export * from "./IContext";
 export * from "./IFetchOptions";
 export * from "./IGraphRequestCallback";
-export * from "./ImplicitMSALAuthenticationProvider";
```
baywet, line 43: This change would be breaking to some people, depending on their import style. Are we planning to major bump?

baywet (changes requested):

versionTest.yml
```diff
@@ -0,0 +1,21 @@
...
+  vmImage: 'ubuntu-latest'
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
+  displayName: 'Install Node.js'
+
+- script: |
+    npm install
```
baywet:
```suggestion
    npm ci
```
this will be faster for CI, it only installs the bare minimum

package.json
```diff
@@ -35,10 +35,13 @@
 		"@types/node": "^12.0.10",
 		"@types/sinon": "^9.0.9",
 		"chai": "^4.2.0",
+		"esm": "^3.2.25",
 		"form-data": "^2.3.3",
+		"fs": "0.0.1-security",
```
baywet, line 40: should be removed as it's not a real package https://www.npmjs.com/package/fs

baywet (changes requested):

versionTest.yml
nikithauc: This file was accidentally added. Not a part of this PR.

docs/ImplicitMSALAuthenticationProvider.md
nikithauc: --save is no longer needed. 
https://stackoverflow.com/questions/19578796/what-is-the-save-option-for-npm-install

README.md
nikithauc: No the authentication flows  are different.

.gitignore
nikithauc: authProviderOptions path contains MD files

spec/core/GraphResponseHandler.ts
nikithauc: This PR is dependent on the test structuring.

src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider.ts
nikithauc: Changed this code.

src/index.ts
nikithauc: It is a major bump.

package.json
nikithauc: We are not exporting in the AuthenticationProviders in the main `src/index.ts`.

.gitignore
ddyett: not clear, why ignore the md files?

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,32 @@
...
+-   Check the [tokenCredentialSamples folder][../samples/tokencredentialsamples]. The ClientSecretCredentialFlow is an example of using TokenCredential to authenticate a node application. The AuthenticationCodeFlow is an example of using TokenCredential to authenticate a browser application.
+
+**Important Note:** TokenCredentials can be used to authentication browser and node applications.
+
+###### Example of how to create and pass a token credential sample -
+
+```typescript
+    // Import the TokenCredential class that you wish to use. This examples uses a Client SecretCredential
+    import { ClientSecretCredential } from "@azure/identity";
+	import { TokenCredentialAuthenticationProvider, TokenCredentialAuthenticationProviderOptions } from "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials";
```
ddyett, line 18: NIT: spacing.

docs/ImplicitMSALAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+		redirectUri: "your_redirect_uri",
+	},
+};
+const graphScopes = ["user.read", "mail.send"]; // An array of graph scopes
+
+// Important Note: This library implements loginPopup and acquireTokenPopup flow, remember this while initializing the msal
+// Initialize the MSAL @see https://github.com/AzureAD/microsoft-authentication-library-for-js#1-instantiate-the-useragentapplication
+const msalApplication = new Msal.UserAgentApplication(msalConfig);
+const options = new MicrosoftGraph.MSALAuthenticationProviderOptions(graphScopes);
+const authProvider = new MicrosoftGraph.ImplicitMSALAuthenticationProvider(msalApplication, options);
```
ddyett, line 25: this looks like the option for how it would be done in javascript not typescript.  Should we show somewhere how to import the implicitmsalauthenticationprovider in typescript?

samples/tokenCredentialSamples/AuthenticationCodeFlow/README.md
```diff
@@ -0,0 +1,22 @@
+# Sample of passing a AuthorizationCodeCredential to the Javascript Client Library
```
ddyett, line 1:
```suggestion
# Sample of passing an AuthorizationCodeCredential to the Javascript Client Library
```

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+ * This sample is referenced from - https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/samples/authorizationCodeSample.ts
+ */
+import { AuthorizationCodeCredential } from "@azure/identity";
+import express from "express";
+import { Server } from "http";
+import "isomorphic-fetch";
+import open from "open";
+import qs from "qs";
+
+import { TokenCredentialAuthenticationProvider } from "../../../../authProviders/azureTokenCredentials";
```
ddyett, line 18: should we just import our own sdk instead so the sample can be standalone from the source.

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+import { Client } from "../../../../lib/src";
+import { port, tenantId, clientSecret, clientId, scopes, authorityHost, redirectUri } from "./secrets";
+
+if (tenantId === undefined || clientId === undefined) {
```
ddyett, line 22: is this all?  i thought you'd need a clientId and scopes as well.

ddyett: can you pass in common for the tenantid?

ddyett:
may also be good to have a pure javascript sample at some point

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
nikithauc: Done. Added the local file path as a dependency.

samples/tokenCredentialSamples/AuthenticationCodeFlow/README.md
nikithauc: Done

package.json
nikithauc: done

docs/ImplicitMSALAuthenticationProvider.md
nikithauc: There is typescript example in this documentation.

.gitignore
```diff
@@ -14,6 +14,14 @@ npm-debug.log
...
+
+src/**/*.js
+src/**/*.js.map
+src/**/*.d.ts
+
+samples/**/*/secrets.ts
+samples/**/*/secrets.js
```
zengin, line 24: Can you give a few sample paths that this covers? Most likely you won't need the last stars.

docs/ImplicitMSALAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
+#### Creating an instance of ImplicitMSALAuthenticationProvider in browser environment
+
+Refer devDependencies in [package.json](../package.json) for the compatible msal version and update that version in below.
+
+**Important Note:** MSAL is supported only for frontend applications, for server-side authentication you have to can use [TokenCredentialAuthenticationProvider](./TokenCredentialAuthenticationProvider.md) or implement your own AuthenticationProvider. Learn how you can create a [Custom Authentication Provider](./docs/CustomAuthenticationProvider.md).
```
zengin, line 5:
```suggestion
**Important Note:** MSAL is supported only for frontend applications, for server-side authentication you can either use [TokenCredentialAuthenticationProvider](./TokenCredentialAuthenticationProvider.md) or implement your own AuthenticationProvider. Learn how you can create a [Custom Authentication Provider](./docs/CustomAuthenticationProvider.md).
```

.gitignore
nikithauc: I am not  clear about this comment.

`authProviderOptions/ ` folder contains the md  file and the index.ts barrels for the submodules.  The `authProviderOptions/ `folder is not git ignored.
This folder is transpiled into the` authProviders/` folder which will be packaged in the npm.
`authProviders/` is ignored in git because it contains the transpiled x.d.ts and  x.js files. 
`authProviders/` is listed in the `package.json` files config so that it is bundled in the JS SDK npm.

samples/tokenCredentialSamples/AuthenticationCodeFlow/package.json
nikithauc: Done

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+
+###### Links for more information -
+
+-   [GitHub - Azure Identity client library for JavaScript ](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/README.md)
+
+-   [npm - Azure Identity client library for JavaScript](https://www.npmjs.com/package/@azure/identity)
+
+-   Check the [tokenCredentialSamples folder][../samples/tokencredentialsamples]. The ClientSecretCredentialFlow is an example of using TokenCredential to authenticate a node application. The AuthenticationCodeFlow is an example of using TokenCredential to authenticate a browser application.
+
+**Important Note:** TokenCredentials can be used to authentication browser and node applications.
```
zengin, line 11:
```suggestion
**Important Note:** TokenCredentials can be used to authenticate browser and node applications.
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+
+###### Example of how to create and pass a token credential sample -
```
zengin, line 13: One of example or sample sounds redundant.

```suggestion
###### Example of how to create and pass a token credential
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+
+```typescript
+// Import the TokenCredential class that you wish to use. This examples uses a Client SecretCredential
```
zengin, line 16:
```suggestion
// Import the TokenCredential class that you wish to use. This example uses a ClientSecretCredential
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+
+import { ClientSecretCredential } from "@azure/identity";
+import { TokenCredentialAuthenticationProvider, TokenCredentialAuthenticationProviderOptions } from "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials";
+
+// Create an instance of the TokenCredential Class that is imported
```
zengin, line 21:
```suggestion
// Create an instance of the TokenCredential class that is imported
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+const tokenCredential = new ClientSecretCredential("your_tenantId", "your_clientId", "your_clientSecret");
+
+// Set your scopes and options for TokenCredential.getToken (Check the ` interface GetTokenOptions` in (TokenCredential Implementation)[https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/core/core-auth/src/tokenCredential.ts])
+
+const options: TokenCredentialAuthenticationProviderOptions = { scopes: [scopes], getTokenoptions };
```
zengin, line 26: Is camel casing only wrong here in the MD file or also in the code?

```suggestion
const options: TokenCredentialAuthenticationProviderOptions = { scopes: [scopes], getTokenOptions };
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+const authProvider = new TokenCredentialAuthenticationProvider(tokenCredential, options);
+const client = Client.initWithMiddleware({
+	debugLogging: true,
+	authProvider: authProvider,
+});
+const res = await client.api("/users/").get();
+```
+
+```javascript
+// Import the TokenCredential class that you wish to use. This examples uses a Client SecretCredential
```
zengin, line 38:
```suggestion
// Import the TokenCredential class that you wish to use. This example uses a ClientSecretCredential
```

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -0,0 +1,56 @@
...
+
+const { ClientSecretCredential } = require("@azure/identity");
+const { Client } = require("@microsoft/microsoft-graph-client");
+const { TokenCredentialAuthenticationProvider } = require("@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials");
+
+// Create an instance of the TokenCredential Class that is imported
```
zengin, line 44:
```suggestion
// Create an instance of the TokenCredential class that is imported
```

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
nikithauc: This is a sample from  
https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/samples/manual/authorizationCodeSample.ts

Can you please explain what does common for the tenant id mean?

samples/tokenCredentialSamples/AuthenticationCodeFlow/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. Navigate to project home [../../]
+
+2. Run `npm install` to install the application.
+
+3. Run `npm run build` to build the library files.
+
+4. Navigate to authenticationCodeFlow samples directory [./samples/tokenCredentialSamples/authenticationCodeFlow]
+
+5.Run `npm start` to build and start the application.
```
zengin, line 22:
```suggestion
5. Run `npm start` to build and start the application.
```

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+			}
+		});
+
+		server = app.listen(port, () => console.log(`Authorization code redirect server listening on port ${port}`));
+	});
+
+	// Direct the user to the authentication URI either by opening a
+	// browser (desktop and mobile apps) or redirecting their browser
+	// using a Location header (web apps and APIs).
+	const authenticateUrl = getAuthorizeUrl(tenantId!, clientId!, scopes);
```
zengin, line 67: naming consistency: please call this variable `authorizeUrl`

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+	console.log("Opening user's browser to URL:", authenticateUrl);
+	await open(authenticateUrl);
+
+	// Wait for the authorization response to be send to the redirect URI
+	const authorizationCode = await authCodePromise;
+	console.log("\nReceived authorization code:", authorizationCode);
+
+	// Once we have the authorization code, the AuthorizationCodeCredential
+	// can be created.  This credential will take care of requesting and
```
zengin, line 76:
```suggestion
	// can be created. This credential will take care of requesting and
```

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+	if (clientSecret) {
+		return new AuthorizationCodeCredential(
+			tenantId!,
+			clientId!,
+			clientSecret,
+			authorizationCode,
+			redirectUri,
+			// NOTE: It is not necessary to explicitly pass the authorityHost when using
+			// the default authority host: https://login.microsoftonline.com.  It is only
+			// necesary when a different authority host is used in the initial authorization
```
zengin, line 87:
```suggestion
			// necessary when a different authority host is used in the initial authorization
```

samples/tokenCredentialSamples/ClientCredentialFlow/README.md
```diff
@@ -0,0 +1,25 @@
+# Sample of passing a ClientSecretCredential to the Javascript Client Library
+
+1. Register your application to enable authentication to Azure Active Directory using the client secret credential. For more details, please check the following links
+
+    - [Register an application](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app)
+
+    - [documention](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis#add-credentials-to-your-web-application).
```
zengin, line 7:
```suggestion
    - [Documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis#add-credentials-to-your-web-application).
```

samples/tokenCredentialSamples/ClientCredentialFlow/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+2. Rename [secrets.sample.ts](./typescript/secrets.sample.ts) to secrets.ts or rename [secrets.sample.js](./javascript/secrets.sample.js) to secrets.js
+
+3. Update your clientId, tenantId, scopes, clientSecret in the **secrets.ts/secrets.js)** file.
```
zengin, line 11:
```suggestion
3. Update your clientId, tenantId, scopes, clientSecret in the **secrets.ts/secrets.js** file.
```

samples/tokenCredentialSamples/ClientCredentialFlow/typescript/secrets.sample.ts
```diff
@@ -0,0 +1,4 @@
+export const tenantId = "<TENANT_ID>";
+export const clientId = "<CLIENT_ID>";
+export const clientSecret = "<CLIENT_SECRET>";
+export const scopes = "<SCOPE>";
```
zengin, line 4: What happens when one needs more than one scope? What type of delimiter can they use?

zengin: From other samples, it looks like this needs to be an array. Sample data implies that it is a string for a single scope.

zengin (changes requested):
Sorry for nitpicking on the typos. I added those comments as most of them appear in public documentation (MD files).

samples/tokenCredentialSamples/ClientCredentialFlow/typescript/secrets.sample.ts
nikithauc: scopes could be an array or a URL that is a string. For example - scope while using the sample would be - "https://graph.microsoft.com/.default".  The secrets.sample files should be renamed and the placeholders should be replaced with the values when someone wants to run the sample.

nikithauc:
> Sorry for nitpicking on the typos. I added those comments as most of them appear in public documentation (MD files).

Thanks for point those out. I should be careful about that.

.gitignore
nikithauc: Corrected the glob to samples/**/secrets.x. 
The secret.sample files should be renamed to -> example - secrets.ts and the sample code imports {credentials} from '/secrets.file'. The secrets.sample files are dummy files with placeholders which should be renamed and replaced by someone checking out the sample. The .gitignore avoids a git push of a file with the actual credentials values.

nikithauc:
> may also be good to have a pure javascript sample at some point

Added.

.gitignore
baywet: oooh ok, I got confused between the two, sorry about that and thanks for detailing!

baywet: nit: conventions are usually to put everything that's going to be in the npm package but not on the repo in a lib or dist folder

nikithauc: Yes. This approach that is used because of the limitations listed in the description.  This can be updated once the new 'exports' field in the package.json -> https://github.com/jkrems/proposal-pkg-exports/  is supported or typescript has a defined of doing this.

samples/tokenCredentialSamples/AuthenticationCodeFlow/src/index.ts
nikithauc: >  @param tenantId - The Azure Active Directory tenant (directory) ID or name.
                   'common' may be used when dealing with multi-tenant scenarios.

https://github.com/Azure/azure-sdk-for-js/blob/9651b096b80d7e437a92266d2f5d16f75c9eb03a/sdk/identity/identity/src/credentials/authorizationCodeCredential.ts#L74

ddyett (approved):

docs/ImplicitMSALAuthenticationProvider.md
baywet: oh I didn't know. Thanks, I guess I'm old :)

package.json
baywet: sure but it's in our module resolution path at runtime, and our SDK will fail at runtime if the dependency is not present.

.gitignore
baywet: what about putting it under lib/dist but without re-exporting it in the main barrel? This way we get a submodule (and it's msal/az-id deps) that's not necessarily in the tree at runtime (only if consumers explicitly take a dependency on it), we're aligned with conventions (making it work with older versions of typescript, not being dependent on a spec that's not implemented yet) and the pipeline/tooling configuration is simpler.

nikithauc: The source is transpiled into the lib directory. We are not changing that. A user can import in both ways - 
1. `const {TokenCredentialAuthenticationProvider} = require("@microsoft/microsoft-graph-client/lib/src/authentication/azureTokenCredentials/TokenCredentialAuthenticationProvider");`
2. `const {TokenCredentialAuthenticationProvider} = require("@microsoft/microsoft-graph-client/authProvider/azureTokenCredentials/");`

Both work.
I considered building into the submodule in the lib folder, but I didn't see much difference between the two paths. 

I don't think there is a strict convention that everything should be inside the lib folder.

Some examples where multiple barrels are used to export in a submodule structure.
- https://github.com/facebook/react/tree/master/packages/react-dom - Example - react-dom/server
- https://github.com/wpengine/headless-framework/tree/main/packages/headless

package.json
nikithauc: Can you give an example of a scenario when you think it will fail?
I created a  local npm package using npm pack script and tried running this example -
// index.js
```
const {Client} = require("@microsoft/microsoft-graph-client");
require("cross-fetch/polyfill");
const client = Client.init({
	defaultVersion: "v1.0",
	debugLogging: true,
	authProvider: (done) => {
		done(null, accessToken);
	},
});
client
	.api("/me")
	.select("displayName")
	.get()
	.then((res) => {
		console.log(res);
	})
	.catch((err) => {
		console.log(err);
	});
```

//package.json dependencies
```
 "dependencies": {
    "@microsoft/microsoft-graph-client": "file:microsoft-microsoft-graph-client-2.1.0-Preview.2.tgz",
    "cross-fetch": "^3.0.6"
  }
```

It worked fine without taking the `azure/identity` as a dependency.
In JS SDK versions < 2.1, we didn't add msal as a dependency and didn't export the ImplicitMSALAuthProviders in the index.ts for the same reason. 

Please let me know if there is something else that can fail this.

README.md
nikithauc: @ddyett @darrelmiller  would know better about this.

docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -8,22 +8,22 @@
...
 import { TokenCredentialAuthenticationProvider, TokenCredentialAuthenticationProviderOptions } from "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials";
 
-// Create an instance of the TokenCredential Class that is imported
+// Create an instance of the TokenCredential class that is imported
 const tokenCredential = new ClientSecretCredential("your_tenantId", "your_clientId", "your_clientSecret");
 
 // Set your scopes and options for TokenCredential.getToken (Check the ` interface GetTokenOptions` in (TokenCredential Implementation)[https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/core/core-auth/src/tokenCredential.ts])
 
-const options: TokenCredentialAuthenticationProviderOptions = { scopes: [scopes], getTokenoptions };
+const options: TokenCredentialAuthenticationProviderOptions = { scopes: [scopes], getTokenOtions };
```
zengin, line 26:
```suggestion
const options: TokenCredentialAuthenticationProviderOptions = { scopes: [scopes], getTokenOptions };
```

samples/tokenCredentialSamples/ClientCredentialFlow/README.md
```diff
@@ -4,11 +4,11 @@
 
     - [Register an application](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app)
 
-    - [documention](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis#add-credentials-to-your-web-application).
+    - [Documention](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis#add-credentials-to-your-web-application).
```
zengin, line 7:
```suggestion
    - [Documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis#add-credentials-to-your-web-application).
```

zengin (approved):

.gitignore
baywet: yeah, because we don't want barrels (to avoid taking hard dependencies) the lib/src part of the first "way of referencing it" cannot be shortened by convention and following the main directive in the package.json... I see.
Should we consider putting the auth provider in a separate package all together? This would resolve:
- keeping this package slim
- having the id packages as runtime dependencies
- having a clear import path
The only drawback is that people will need to install a second package, but that's not a big problem in this ecosystem as long as we document it.

package.json
baywet: in your example here, just add 
```JS
const {TokenCredentialAuthenticationProvider} = require("@microsoft/microsoft-graph-client/authProvider/azureTokenCredentials/");
```
at the top.
and assign a new provider to the client, without installing the dep.

nikithauc: "peerDependenciesMeta": {
		"msal": {
			"optional": true
		},
		"@azure/identity": {
			"optional": true
		}
	}

This indicates that  '@azure/identity' is an optional peer dependency .
I think this field was lost while merging changes. I readded this.

.gitignore
nikithauc: This point has been discussed and there is no substantial scope or functionality for a separate auth library for JS. 
If we are dependent on multiple authentication libraries a separate library would be a feasible option. With TokenCredentials, the goal would be to provide a single easy interface for a variety of authentication flows and keep the customAuthProvider implementation small.

README.md
nikithauc: https://github.com/Azure/azure-sdk-for-js/pull/13263
https://github.com/Azure/azure-sdk-for-js/issues/12565#issuecomment-732429468
My knowledge is that working with a single dependency that is @azure/identity which provides a lot of authentication flows (and plans to bring in more auth flows) would reduce work for us keeping the implementation short rather than taking separate dependencies or multiple implementations to support node or browser auth.

package.json
baywet: I think this is an improvement as it'll trigger a warning during the install if the current "package" doesn't have it in it's dependency tree. People might still miss the warning though. And people who don't need that dep (because they are not using the auth provider) will get confused because of the warning.

.gitignore
baywet: alright. Thanks for taking the time to discuss this. We'll see what feedback we get from the community once we ship :)

package.json
nikithauc: The optional field when set true does not trigger a warning if the peer is not present. This caters our requirement. 
If the user doesn't need to use '@azure/identity', he will not be warned. We have an indicator  in the package.json that says we need the optional dependency if the user is using the TokenCredentialAuthProvider. Runtime failure is expected for someone using the TokenCredentialAuthProvider without installing the '@azure/identity' because they shouldn't be using the TokenCredentialAuthProvider class if the TokenCredentials are not used for authentication. 

> If true, the selected peer dependency will be marked as optional by the package manager and the consumer omitting it won't be reported as an error. 
https://yarnpkg.com/configuration/manifest#peerDependenciesMeta.optional

README.md
baywet: ok. Thanks for looking into that further. I was confused about why they'd "mask" capabilities they get for "free" from MSAL that customers might need. But hopefully, in the fullness of time, this current PR adding azure identity adds support for a whole bunch of flows (browser, backend, etc...). Resolving

package.json
baywet: and if they need the `TokenCredentialAuthenticationProvider`, they will have to pass in a `TokenCredential` which comes from azure identity forcing them to install the needed dep. Makes sense.
Now last question: do we need msal listed out there as it's a dep from azure identity?

nikithauc: If someone uses ImplicitMsalAuthProvider, they will need msal 1.0 . 
msal is not listed as a runtime dependency anymore. It is an optional dependency just like @azure/identity. The exports for the ImplicitMsalAuthProvider have been removed from the main barrel src/index.ts.  The exports were added only in version 2.1 which caused need for adding msal as a runtime dependency.

baywet: Great. Thanks for walking through the discussion with me!. Resolving

baywet (approved):

Stono:
Hey folks
I'm wondering if you have an ETA on 3.0.0 as I need this PR :) 
Thanks
KArl

PR #376, opened 2021-01-12
Update Typescript version to 4.x.

nikithauc:


baywet (approved):
The CI is failing for node 7.
I think we should drop anything below node 10 from the CI and maybe add versions 12, 14 and 16 to follow the [support matrix](https://nodejs.org/en/about/releases/)

ddyett (approved):

PR #389, opened 2021-02-02
Migration to ESLint

nikithauc:
Summary of changes 

1. Added eslint rules, picking some rules from https://github.com/microsoft/TypeScript/blob/master/.eslintrc.json.
2. Removed try catches where the catch was simply rethrowing the error.
3. Applied [ no-inferrable-types](https://palantir.github.io/tslint/rules/no-inferrable-types/).  Example - `const x:number = 1` will now be `const x = 1`. The goal of this to avoid redundancy and verbose code.
4. Sorted imports.
5. Replaced `options.hasOwnProperty(i)` with `Object.prototype.hasOwnProperty.call(options, i)`. https://stackoverflow.com/questions/12017693/why-use-object-prototype-hasownproperty-callmyobj-prop-instead-of-myobj-hasow
6. Additional code  improvements with use of variables.


.eslintignore
```diff
@@ -0,0 +1,8 @@
+*.js
+*.js.map
+*.d.ts
+
+node_modules
+lib
+samples
```
baywet, line 7: maybe we should also lint our samples? So we present "pretty" code.

.eslintrc.json
```diff
@@ -0,0 +1,192 @@
+{
+	"root": true,
+	"parser": "@typescript-eslint/parser",
+	"plugins": ["@typescript-eslint", "prettier", "simple-import-sort"],
+	"extends": ["eslint:recommended", "plugin:@typescript-eslint/eslint-recommended", "plugin:@typescript-eslint/recommended", "prettier"],
+	"rules": {
+		"no-useless-catch": "error", //TODO- Remove the unnecessary try-catch lines used and enable no-useless-catch rule.
```
baywet, line 7: nit: technically .json doesn't accept comments. jsonc does. But I'm not sure whether eslint accepts it.

src/Client.ts
```diff
@@ -45,7 +45,7 @@ export class Client {
 	public static init(options: Options): Client {
 		const clientOptions: ClientOptions = {};
 		for (const i in options) {
-			if (options.hasOwnProperty(i)) {
+			if (Object.prototype.hasOwnProperty.call(options, i)) {
```
baywet, line 48: I'm not sure which rule triggered that change but we should probably tweak the settings. This makes the code convoluted. My understanding is that it's done because JS doesn't guard against using the "hasOwnProperty" property name and this prevents using a local definition. But we shouldn't need that. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name

src/GraphRequest.ts
```diff
@@ -631,12 +629,8 @@ export class GraphRequest {
 		const options: FetchOptions = {
 			method: RequestMethod.GET,
 		};
-		try {
-			const response = await this.send(url, options, callback);
-			return response;
-		} catch (error) {
-			throw error;
-		}
+		const response = await this.send(url, options, callback);
```
baywet, line 632: return the response directly? (x5/6)

src/middleware/HTTPMessageHandler.ts
```diff
@@ -27,11 +26,7 @@ export class HTTPMessageHandler implements Middleware {
...
 	 */
 	public async execute(context: Context): Promise<void> {
-		try {
-			context.response = await fetch(context.request, context.options);
-			return;
-		} catch (error) {
-			throw error;
-		}
+		context.response = await fetch(context.request, context.options);
+		return;
```
baywet, line 30: I don't believe the return is required here

baywet (changes requested):
Thanks for making that change. This is greatly going to reduce review workload as well as bring some extra consistency to the code.

src/GraphRequest.ts
MIchaelMainer: Vincent's comment is applicable to lines 658, 689, 708, 736, 764, and 785

src/Client.ts
nikithauc: This rule is recommended because of possible vulnerabilities -  
https://eslint.org/docs/rules/no-prototype-builtins
https://www.nickang.com/2018-02-04-developers-use-object-prototype-hasownproperty-call/

The code looks verbose but avoids the certain vulnerabilities.

.eslintignore
nikithauc: Done

src/GraphRequest.ts
nikithauc: Done

src/middleware/HTTPMessageHandler.ts
nikithauc: Done

.eslintrc.json
nikithauc: Removed the comments. I added few comments to separate the type of rules. Removed those as well.

src/Client.ts
baywet: I understand the argument of the method being defined locally to something else. Or the one about creating objects without prototypes. As long as we use hasOwnProperty on objects we created, and as long as we don't do either of those, we should be good on that front.
I believe the Map attack argument is a moot point now as we ship to es2015/es6 and that's been fixed.
But better be safe than sorry. Thanks for the context.

baywet (approved):

src/content/BatchRequestContent.ts
```diff
@@ -227,14 +227,14 @@ export class BatchRequestContent {
...
-		let bodyParsed: boolean = false;
+		let bodyParsed = false;
 		let body;
 		try {
 			const cloneReq = request.clone();
 			body = await cloneReq.json();
 			bodyParsed = true;
 		} catch (e) {
-			// tslint:disable-line: no-empty
+			//TODO- Handle empty catches
```
zengin, line 237: Please create an issue for this and link it here.

src/middleware/options/TelemetryHandlerOptions.ts
```diff
@@ -21,11 +20,12 @@ import { MiddlewareOptions } from "./IMiddlewareOptions";
  * @property {number} RETRY_HANDLER_ENABLED - The hexadecimal flag value for retry handler enabled
  * @property {number} AUTHENTICATION_HANDLER_ENABLED - The hexadecimal flag value for the authentication handler enabled
  */
+
 export enum FeatureUsageFlag {
-	NONE = 0x0,
-	REDIRECT_HANDLER_ENABLED = 0x1,
-	RETRY_HANDLER_ENABLED = 0x2,
-	AUTHENTICATION_HANDLER_ENABLED = 0x4,
+	NONE = 0x0, // eslint-disable-line @typescript-eslint/naming-convention
```
zengin, line 25: Is there a way to define the ignore rule for multiple lines similar to C#?

test/common/core/Client.ts
```diff
@@ -46,7 +47,7 @@ describe("Client.ts", () => {
 			};
 			const client: Client = Client.initWithMiddleware(options);
 			assert.isTrue(client instanceof Client);
-			assert.isDefined(client["httpClient"]);
+			assert.isDefined(["httpClient"]);
```
zengin, line 50: how does this work?

zengin (changes requested):

test/common/core/Client.ts
nikithauc: Thanks for catching this! Corrected this.

src/middleware/options/TelemetryHandlerOptions.ts
nikithauc: Replaced with a single line.

src/content/BatchRequestContent.ts
nikithauc: https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/390

src/middleware/options/TelemetryHandlerOptions.ts
```diff
@@ -22,10 +22,11 @@ import { MiddlewareOptions } from "./IMiddlewareOptions";
...
 export enum FeatureUsageFlag {
-	NONE = 0x0, // eslint-disable-line @typescript-eslint/naming-convention
-	REDIRECT_HANDLER_ENABLED = 0x1, // eslint-disable-line @typescript-eslint/naming-convention
-	RETRY_HANDLER_ENABLED = 0x2, // eslint-disable-line @typescript-eslint/naming-convention
-	AUTHENTICATION_HANDLER_ENABLED = 0x4, // eslint-disable-line @typescript-eslint/naming-convention
+	/* eslint-disable  @typescript-eslint/naming-convention */
+	NONE = 0x0,
+	REDIRECT_HANDLER_ENABLED = 0x1,
+	RETRY_HANDLER_ENABLED = 0x2,
+	AUTHENTICATION_HANDLER_ENABLED = 0x4,
```
zengin: I think you need to reenable to rule, otherwise it is for the entire file from line 25.
```suggestion
	AUTHENTICATION_HANDLER_ENABLED = 0x4,
	/* eslint-enable  @typescript-eslint/naming-convention */
```

zengin (changes requested):

src/middleware/options/TelemetryHandlerOptions.ts
nikithauc: Done.

zengin (approved):

Issue #394, opened 2021-02-18
Use browser field in package.json, remove graph-es-sdk.js file.


nikithauc:
> The browser field is provided by a module author as a hint to javascript bundlers or component tools when packaging modules for client side use. The field is found in a package.json file (described here) usually located at the root of a project source tree.

https://github.com/defunctzombie/package-browser-field-spec
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#browser

`msgraph-sdk-javascript` is an isomorphic library. 

Use the browser field for the following - 
1. We currently have two entry files, src/index.ts and src/browser/index.ts. 
 - Use the browser field to indicate the browser entry point. 
Example - 
 ```
"browser":
{ "lib/es/index.node.js": "lib/es/index.browser.js" }
```
- Currently, we have  only "module": "lib/es/index.js" in the package.json. This will remain the same. 
2. Better way to handle environment specific implementation. For example, using the browser field we can indicate as follows - 
```
"browser":
{ "lib/es/featureNode.js": "lib/es/featureBrowser.js" }
```
The browser specific implementation should be exported in the browser entry point.

3. We currently use rollup.js to bundle our library into IIFE and ES formats for browser use. The [src/browser/index.ts](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/src/browser/index.ts) is the entry point for rollup.js
   - The process of bundling library in IIFE format should continue with some updates.
   - I suggest we can remove the `graph-es-sdk.js` as we are shipping the ES modules. Currently, the entry points for the bundled `graph-es-sdk.js` file and the ES modules are different. We should have one source for the ES modules and the user can bundle as per their requirement.
     
[AB#8194](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8194)

PR #396, opened 2021-03-01
Nikithauc/publish doc

nikithauc:
#394

### DIFF
--- a/design/publishing.md
+++ b/design/publishing.md
@@ -1,18 +1,20 @@
 ##### This design document focuses on the following - 
 1. Separate entry points for node and browser.
-2. Specifying the browser field in package.json
-3. Changes in the bundling process
+2. Specifying the browser field in package.json.
+3. Changes in the bundling process.
 
 ##### Terms -
 
-bundler - Module bundlers are tools frontend developers used to bundle JavaScript modules into a single JavaScript files that can be executed in the browser.
+* bundler - Module bundlers are tools frontend developers used to bundle JavaScript modules into a single JavaScript files that can be executed in the browser.
 
-package.json fields - 
-* main - The main field is a module ID that is the primary entry point to your program. Points to the CJS modules.
+* rollup - rollup.js is the module bundler that the JS SDK uses. 
 
-* module - The module field is not an official npm feature but a common convention among bundlers to designate how to import an ESM version of a library. Points to the ES modules.
+* package.json fields - 
+  * main - The main field is a module ID that is the primary entry point to your program. Points to the CJS modules.
 
-* browser - If your module is meant to be used client-side the browser field should be used instead of the main field.
+  * module - The module field is not an official npm feature but a common convention among bundlers to designate how to import an ESM version of a library. Points to the ES modules.
+
+  * browser - If the module is meant to be used client-side, the browser field should be used instead of the main field.
 
 ##### Current set up -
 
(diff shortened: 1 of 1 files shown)


design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+                      Transpiles into JavaScript 
+                             'lib' folder
+                               /   \
+                     CJS module     ES modules
+2. main - `lib/src/index.js`
+   module - `lib/es/src/index.js`
+
+3. Rollup bundling output
+* `graph-js-sdk.js` - Bundled and minified file in IIFE format. This file can be directly used in the browser with a `<script>` tag.
+* `graph-es-sdk.js` - Bundled filein ES format.
```
baywet, line 33:
```suggestion
* `graph-es-sdk.js` - Bundled file in ES format.
```

design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+`src/ImplicitMsalProvider` imports or requires 'msal' in the implementation. 
+4. My assumtion is that `src/browser/ImplicitMsalProvider` is implemented specifically for the rollup process and to skip the rollup external dependency while using `graph-js-sdk.js` in the browser.
+
+Note - Browser applications using the ES modules from the npm package of the JS SDK refer to the `module` entry point - `lib/es/src/index.js`(not the browser entry point). Example - Graph Explorer.
+
+##### Upcoming changes - 
+
+1. Use the browser field for the following -
+
+* We currently have two entry files, `src/index` and `src/browser/index`.
```
baywet, line 49: nit: who's "we" usually the docs people tell me to avoid using ambiguous subjects like those and clarify by using "Microsoft", or some other more specific term.

design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+Use the browser field to indicate the browser entry point.
+Example -
+"browser":
+{ "lib/es/src/index.js": "lib/es/src/browser/index.js" }
```
baywet, line 53: nit: format as json

design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+* `graph-js-sdk.js` - Bundled and minified file in IIFE format. This file can be directly used in the browser with a `<script>` tag.
+* `graph-es-sdk.js` - Bundled filein ES format.
+4. Entry point for rollup - `lib/es/src/browser/index.js`.
+
+##### Difference between src/index.js and src/browser/index.js
+1. `src/browser/index.js` does not export  `RedirectHandler` and `RedirectHandlerOptions`. Redirection is handled by the browser.
+2. `src/browser/index.js` exports `src/browser/ImplicitMsalProvider`.
+3. `src/browser/ImplicitMsalProvider` does not import or require 'msal' dependency. While, 
+`src/ImplicitMsalProvider` imports or requires 'msal' in the implementation. 
+4. My assumtion is that `src/browser/ImplicitMsalProvider` is implemented specifically for the rollup process and to skip the rollup external dependency while using `graph-js-sdk.js` in the browser.
```
baywet, line 41: we shouldn't need both. The MSAL dependency should be identified as an external dependency by the bundler and referenced but not included in the resulting bundle. If this is not the case today, there's probably a configuration with our bundler.

design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+Currently, we have the main and "module field in the package.json. This will remain the same.
+* Better way to handle environment specific implementation. For example, using the `browser` field we can indicate as follows -
+"browser":
+{ "stream": "stream-browserify",
+  "Feature-Node.js": "Feature-Browser.js"
+} 
+
+
+2. Remove  export of `src/browser/ImplicitMsalProvider` from `src/browser/index`.
+* Till `ImplicitMsalProvider` is maintained in the repo, maintain a separate entry point say `rollup-index` for the rollup process which exports  `src/browser/index` and `src/browser/ImplicitMsalProvider`.
```
baywet, line 63: is the plan to completely remove the implicit provider in the long term because it's not recommended from a security standpoint?

design/publishing.md
```diff
@@ -0,0 +1,69 @@
...
+  "Feature-Node.js": "Feature-Browser.js"
+} 
+
+
+2. Remove  export of `src/browser/ImplicitMsalProvider` from `src/browser/index`.
+* Till `ImplicitMsalProvider` is maintained in the repo, maintain a separate entry point say `rollup-index` for the rollup process which exports  `src/browser/index` and `src/browser/ImplicitMsalProvider`. 
+* Continue rolling up the `src/browser/ImplicitMsalProvider` as we currently do and not introduce breaking changes here as we are planning to deprecate this.
+* Remove the separate entry point once `ImplicitMsalProvider` is removed and use the browser entry point for roll up thereafter. The goal is to maintain a consistent entry point or usage for browser applications using the JS SDK and the rollup/bundling process. 
+
+3. Bundle the authproviders separately as they are optional.
```
baywet, line 67: do you also intent to ship different packages?

design/publishing.md
nikithauc: To be discussed in the Wednesday meeting.

design/publishing.md
nikithauc: Can you point to the configuration in the bundler? 
 
An external dependency should be marked in the `rollup.config`  file which is not done currently. I agree that dependency should be noted in the roll up process going forward to maintain consistency between the modules and the bundled files.

design/publishing.md
nikithauc: No. Just maintain a separate entry point in the `rollup.config` with external dependency specified only for the authprovider bundles.  Example, `graph-auth-tokenCred.js` which has the `@azure/identity` dependency marked as external. The external dependency cannot be marked optional in case of the rollup process.

design/publishing.md
baywet: I'm not super familiar with rollup but [this](https://stackoverflow.com/a/41432115/3808675) seems to be the way to do it.

design/publishing.md
nikithauc: Yes.

design/publishing.md
nikithauc: I assumed you were talking about a config in our bundler set up, something that I might have missed.

baywet: webpack has an externals configuration but we're not using webpack :)

baywet (approved):

nmetulev (approved):

PR #397, opened 2021-03-01
Large file upload task - Add UploadResult object

### LINKED ISSUE #359: LargeFileUploadTask is ending with PassThrough stream instead of saying JSON response with id
# Bug Report
I'm using  msgraph-sdk-javascript in NodeJs and try attach a document in Outlook email. The issue is, LargeFileUploadTask.uploadSlice function returns PassThrough readable stream on end of file upload instead of JSON response with id.

## Prerequisites
-   [ Yes] Can you reproduce the problem?
-   [Yes ] Are you running the latest version?
I'm using 2.1.1 as of 19th Nov 2020
-   [Yes ] Are you reporting to the correct repository?
-   [ ] Did you perform a cursory search?

## Steps to Reproduce
1. Create an Outlook draft message by usingPostman. 

2. Create a simple HttpTrigger Azure Function and Add following code in Azure Function handler.
```
const saveAttachment = async (oid, tenantId, messageId, file) => {
    // getClient is just creating client with GraphToken generate authProvider.
    // authProvider is doing client-credential flow to generate token.
    const client = await getClient(tenantId);

    const uploadSessionPayload = {
        AttachmentItem: {
          attachmentType: 'file',
          name: file.filename,
          size: file.size
        }
    };

    const uploadSession = await LargeFileUploadTask.createUploadSession(
        client, 
        `/users/${oid}/messages/${messageId}/attachments/createUploadSession`,
        uploadSessionPayload,
    );
    const fileObject = {
        content: file.data,
        name: file.filename,
        size: file.size
    };

    const { url: uploadUrl } = uploadSession; 
    const uploadUrlTokens = uploadUrl.split('?');
    const uploadUrlAuthToken = uploadUrlTokens[1].replace('authtoken=', '');
    const uploadClient = Client.init({
        authProvider: (done) => done(null, uploadUrlAuthToken),
        debugLogging: true
    });

    const uploadTask = new LargeFileUploadTask(uploadClient, fileObject, uploadSession, {
        rangeSize: 5 * 1024 * 1024,
    });

    const uploadResponse = await uploadTask.upload();
    return uploadResponse;
};
```
3. Execute AzureFunction in Visual Studio Code.

4. Call HttpTrigger from Postman

**Expected behavior:**
Large file should be uploaded in sequential PUT call. In end of all chunks upload, `LargeFileUploadTask.uploadSlice` function should give response with 'id'

ShouldEndUpFileJsonResponseWithIdProperty

**Actual behavior:** 
`LargeFileUploadTask.uploadSlice` function is giving NodeJs ReadableStream.

## Additional Context
LargeFileUploadResponse1

LargeFileUploadResponseEndOfUpload

Code Line:

https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/41306143d8978f5256b20354adf7486451a11dc6/src/tasks/LargeFileUploadTask.ts#L254
 
[AB#7549](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7549)

nikithauc:
Fixes #359. 

* Change the condition to mark upload as completed.
* Added a `UploadResult` class.
* Added a `getUploadSession()` function.
* Added `isCancelled` property to `LargeUploadSession` interface.

### DIFF
--- a/docs/tasks/LargeFileUploadTask.md
+++ b/docs/tasks/LargeFileUploadTask.md
@@ -129,3 +129,19 @@ async function largeFileUpload(client, file) {
 	}
 }
 ```
+
+## Cancelling a largeFileUpload task
+
+_Cancelling an upload session sends a DELETE request to the upload session URL_
+
+```typescript
+const cancelResponse = await uploadTask.cancel();
+```
+
+## Get the largeFileUpload session
+
+_Returns the largeFileUpload session information containing the URL, expiry date and cancellation status of the task_
+
+```typescript
+const uploadsession: LargeFileUploadSession = uploadTask.getUploadSession();
+```
--- a/src/middleware/AuthenticationHandler.ts
+++ b/src/middleware/AuthenticationHandler.ts
@@ -9,6 +9,7 @@
  * @module AuthenticationHandler
  */
 
+import { isGraphURL } from "../GraphRequestUtil";
 import { AuthenticationProvider } from "../IAuthenticationProvider";
 import { AuthenticationProviderOptions } from "../IAuthenticationProviderOptions";
 import { Context } from "../IContext";
@@ -60,6 +61,8 @@ export class AuthenticationHandler implements Middleware {
 	 * @returns A Promise that resolves to nothing
 	 */
 	public async execute(context: Context): Promise<void> {
+		const url = typeof context.request === "string" ? context.request : context.request.url;
+			if (isGraphURL(url)) {
 		let options: AuthenticationHandlerOptions;
 		if (context.middlewareControl instanceof MiddlewareControl) {
 			options = context.middlewareControl.getMiddlewareOptions(AuthenticationHandlerOptions) as AuthenticationHandlerOptions;
@@ -77,6 +80,11 @@ export class AuthenticationHandler implements Middleware {
 		const bearerKey = `Bearer ${token}`;
 		appendRequestHeader(context.request, context.options, AuthenticationHandler.AUTHORIZATION_HEADER, bearerKey);
 		TelemetryHandlerOptions.updateFeatureUsageFlag(context, FeatureUsageFlag.AUTHENTICATION_HANDLER_ENABLED);
+	} else {
+		if (context.options.headers) {
+			delete context.options.headers[AuthenticationHandler.AUTHORIZATION_HEADER];
+		}
+	}
 		return await this.nextMiddleware.execute(context);
 	}
 
--- a/src/middleware/AuthenticationHandler.ts
+++ b/src/middleware/AuthenticationHandler.ts
@@ -69,11 +69,11 @@ export class AuthenticationHandler implements Middleware {
 			}
 			let authenticationProvider: AuthenticationProvider;
 			let authenticationProviderOptions: AuthenticationProviderOptions;
-			if (typeof options !== "undefined") {
+			if (options) {
 				authenticationProvider = options.authenticationProvider;
 				authenticationProviderOptions = options.authenticationProviderOptions;
 			}
-			if (typeof authenticationProvider === "undefined") {
+			if (!authenticationProvider) {
 				authenticationProvider = this.authenticationProvider;
 			}
 			const token: string = await authenticationProvider.getAccessToken(authenticationProviderOptions);
--- a/src/middleware/AuthenticationHandler.ts
+++ b/src/middleware/AuthenticationHandler.ts
@@ -62,29 +62,29 @@ export class AuthenticationHandler implements Middleware {
 	 */
 	public async execute(context: Context): Promise<void> {
 		const url = typeof context.request === "string" ? context.request : context.request.url;
-			if (isGraphURL(url)) {
-		let options: AuthenticationHandlerOptions;
-		if (context.middlewareControl instanceof MiddlewareControl) {
-			options = context.middlewareControl.getMiddlewareOptions(AuthenticationHandlerOptions) as AuthenticationHandlerOptions;
+		if (isGraphURL(url)) {
+			let options: AuthenticationHandlerOptions;
+			if (context.middlewareControl instanceof MiddlewareControl) {
+				options = context.middlewareControl.getMiddlewareOptions(AuthenticationHandlerOptions) as AuthenticationHandlerOptions;
+			}
+			let authenticationProvider: AuthenticationProvider;
+			let authenticationProviderOptions: AuthenticationProviderOptions;
+			if (typeof options !== "undefined") {
+				authenticationProvider = options.authenticationProvider;
+				authenticationProviderOptions = options.authenticationProviderOptions;
+			}
+			if (typeof authenticationProvider === "undefined") {
+				authenticationProvider = this.authenticationProvider;
+			}
+			const token: string = await authenticationProvider.getAccessToken(authenticationProviderOptions);
+			const bearerKey = `Bearer ${token}`;
+			appendRequestHeader(context.request, context.options, AuthenticationHandler.AUTHORIZATION_HEADER, bearerKey);
+			TelemetryHandlerOptions.updateFeatureUsageFlag(context, FeatureUsageFlag.AUTHENTICATION_HANDLER_ENABLED);
+		} else {
+			if (context.options.headers) {
+				delete context.options.headers[AuthenticationHandler.AUTHORIZATION_HEADER];
+			}
 		}
-		let authenticationProvider: AuthenticationProvider;
-		let authenticationProviderOptions: AuthenticationProviderOptions;
-		if (typeof options !== "undefined") {
-			authenticationProvider = options.authenticationProvider;
-			authenticationProviderOptions = options.authenticationProviderOptions;
-		}
-		if (typeof authenticationProvider === "undefined") {
-			authenticationProvider = this.authenticationProvider;
-		}
-		const token: string = await authenticationProvider.getAccessToken(authenticationProviderOptions);
-		const bearerKey = `Bearer ${token}`;
-		appendRequestHeader(context.request, context.options, AuthenticationHandler.AUTHORIZATION_HEADER, bearerKey);
-		TelemetryHandlerOptions.updateFeatureUsageFlag(context, FeatureUsageFlag.AUTHENTICATION_HANDLER_ENABLED);
-	} else {
-		if (context.options.headers) {
-			delete context.options.headers[AuthenticationHandler.AUTHORIZATION_HEADER];
-		}
-	}
 		return await this.nextMiddleware.execute(context);
 	}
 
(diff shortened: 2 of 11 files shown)


src/middleware/AuthenticationHandler.ts
```diff
@@ -60,23 +61,30 @@ export class AuthenticationHandler implements Middleware {
...
-			options = context.middlewareControl.getMiddlewareOptions(AuthenticationHandlerOptions) as AuthenticationHandlerOptions;
+		const url = typeof context.request === "string" ? context.request : context.request.url;
+		if (isGraphURL(url)) {
+			let options: AuthenticationHandlerOptions;
+			if (context.middlewareControl instanceof MiddlewareControl) {
+				options = context.middlewareControl.getMiddlewareOptions(AuthenticationHandlerOptions) as AuthenticationHandlerOptions;
+			}
+			let authenticationProvider: AuthenticationProvider;
+			let authenticationProviderOptions: AuthenticationProviderOptions;
+			if (typeof options !== "undefined") {
```
baywet, line 72:
```suggestion
			if (options) {
```

src/middleware/AuthenticationHandler.ts
```diff
@@ -60,23 +61,30 @@ export class AuthenticationHandler implements Middleware {
...
+				authenticationProvider = options.authenticationProvider;
+				authenticationProviderOptions = options.authenticationProviderOptions;
+			}
+			if (typeof authenticationProvider === "undefined") {
```
baywet, line 76:
```suggestion
			if (!authenticationProvider) {
```

src/tasks/FileUploadUtil/UploadResult.ts
```diff
@@ -0,0 +1,77 @@
...
+		this._location = location;
+		this._responseBody = responseBody;
+	}
+
+	/**
+	 * @public
+	 * @param {responsebody} responsebody - The response body from the completed upload response
+	 * @param {responsebody} responseHeaders - The headers from the completed upload response
+	 */
+	public static CreateUploadResult(responsebody?: unknown, responseHeaders?: Headers) {
```
baywet, line 74: any specific reason why this is not simply a constructor?

src/tasks/LargeFileUploadTask.ts
```diff
@@ -50,6 +54,7 @@ export interface LargeFileUploadTaskOptions {
 export interface LargeFileUploadSession {
 	url: string;
 	expiry: Date;
+	isCancelled?: boolean;
```
baywet, line 57: why is a nullable needed here? simply use a boolean (non-nullable) which will default to false

src/tasks/LargeFileUploadTask.ts
```diff
@@ -124,6 +129,7 @@ export class LargeFileUploadTask {
 		const largeFileUploadSession: LargeFileUploadSession = {
 			url: session.uploadUrl,
 			expiry: new Date(session.expirationDateTime),
+			isCancelled: false,
```
baywet, line 132: not needed if a non-nullable is used

src/tasks/LargeFileUploadTask.ts
```diff
@@ -259,7 +300,14 @@ export class LargeFileUploadTask {
 	 * @returns The promise resolves to cancelled response
 	 */
 	public async cancel(): Promise<any> {
-		return await this.client.api(this.uploadSession.url).delete();
+		const cancelResponse = await this.client
+			.api(this.uploadSession.url)
+			.responseType(ResponseType.RAW)
+			.delete();
+		if (cancelResponse.status === 204) {
```
baywet, line 307: it'd be nice to document the cancel capability in the design doc, do you mind taking a stab at it please?

baywet (changes requested):

src/tasks/LargeFileUploadTask.ts
nikithauc: > No. You cannot provide default values for interfaces or type aliases as they are compile time only and default values need runtime support
https://stackoverflow.com/questions/35074365/typescript-interface-default-values

Can you please give me an example if you are suggesting some else?

src/tasks/FileUploadUtil/UploadResult.ts
nikithauc: Yes, the intention is to keep the implementation open say if there is more than one way or more than a single source to populate the UploadResult object.

src/tasks/LargeFileUploadTask.ts
nikithauc: Done

src/tasks/FileUploadUtil/UploadResult.ts
baywet: because of the single constructor limitation? What do you think of using a constructor with optional parameters and default values? This way the caller would use a standard way of creating the object, pass-in what they can. Or even better, using a local initialization interface like [described here](https://betterprogramming.pub/named-parameters-in-typescript-e32c763d2b2e) ? (this would effectively add a behavior like named parameters, be extensible, and it's a very common practice in TS)

src/tasks/LargeFileUploadTask.ts
baywet: As far as I understand, you don't need to set that default value. The default of bool (non-nullable) is false. Just the only change here would be to remove the ? sign.

docs/tasks/LargeFileUploadTask.md
```diff
@@ -129,3 +129,19 @@ async function largeFileUpload(client, file) {
 	}
 }
 ```
+
+## Cancelling a largeFileUpload task
```
baywet, line 133: this will need to be surface up here https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/tasks/FileUploadTask.md

src/tasks/LargeFileUploadTask.ts
nikithauc: The point of making this parameter optional is that in most cases the `isCancelled status`  will be false. Removing the ? makes it necessary for the user to specify for the user everytime a `LargeFileUploadSession` is created. The `LargeFileUploadSession` is created and passed when created a large file upload task and the status `isCancelled` will always be false in this case.

baywet: right. Sorry I forgot that in some cases typescript would enforce providing non-nullable properties on objects. I believe one could get around using deconstruct and/or assign but that gets messy.
Thanks for the clarification and sorry about the mumbling here 😂

src/tasks/FileUploadUtil/UploadResult.ts
nikithauc: Yes right I did this because of the single constructor limitation. 

I prefer the static factory pattern here - 

I did consider named parameters or having an interface with default/optional parameters. My thought is that if tomorrow I have to add another way of populating the UploadResult adding the logic in a separate method would be clean and also locating the method will be easier. Something like `CreateUploadResultWithNewAPIResult()` , `CreateUploadResultWithBetaResult()`. The name of the factory method will clarify what we need and its easier to search for a method than go through list of parameters and set what is needed. Also, grouping related parameters for a named structure would be clean. I don't want to group unrelated parameters.

> Consider public static methods which return a new Matrix instead of constructors overloads.
https://codereview.stackexchange.com/questions/124849/emulating-constructor-overloading-in-typescript

docs/tasks/LargeFileUploadTask.md
nikithauc: Working on it.

nikithauc: https://github.com/microsoftgraph/msgraph-sdk-design/pull/43

src/tasks/FileUploadUtil/UploadResult.ts
baywet: Yeah I guess I'm coming at "too much of a react angle" for the interface and objects deconstruct approach.
I can't see an obvious benefit for one or the other besides conventions and preferences :)
last thing, can you change responsebody to responseBody for consistency with the other parameter please?

nikithauc: Changed to responseBody!

src/tasks/FileUploadUtil/UploadResult.ts
```diff
@@ -68,10 +68,10 @@ export class UploadResult {
 
 	/**
 	 * @public
-	 * @param {responsebody} responsebody - The response body from the completed upload response
+	 * @param {responsebody} responseBody - The response body from the completed upload response
```
baywet, line 71:
```suggestion
	 * @param {responseBody} responseBody - The response body from the completed upload response
```

src/tasks/FileUploadUtil/UploadResult.ts
```diff
@@ -0,0 +1,77 @@
...
+	public constructor(responseBody: unknown, location: string) {
+		// Response body or the location parameter can be undefined.
+		this._location = location;
+		this._responseBody = responseBody;
+	}
+
+	/**
+	 * @public
+	 * @param {responsebody} responseBody - The response body from the completed upload response
+	 * @param {responsebody} responseHeaders - The headers from the completed upload response
```
baywet, line 72:
```suggestion
	 * @param {responseHeaders} responseHeaders - The headers from the completed upload response
```

baywet (changes requested):

src/tasks/FileUploadUtil/UploadResult.ts
```diff
@@ -0,0 +1,77 @@
...
+	 * @public
+	 * Set the response body from the completed upload response
+	 */
+	public set responseBody(responseBody: unknown) {
+		this._responseBody = responseBody;
+	}
+
+	/**
+	 * @public
+	 * @param {responsebody} responsebody - The response body from the completed upload response
```
baywet, line 60:
```suggestion
	 * @param {responseBody} responsebody - The response body from the completed upload response
```

src/tasks/FileUploadUtil/UploadResult.ts
```diff
@@ -0,0 +1,77 @@
...
+	 * @param {responsebody} location - The location value from the headers from the completed upload response
```
baywet, line 61:
```suggestion
	 * @param {location} location - The location value from the headers from the completed upload response
```

baywet (changes requested):

baywet (approved):

test/common/tasks/LargeFileUploadTask.ts
```diff
@@ -157,12 +159,66 @@ describe("LargeFileUploadTask.ts", () => {
...
-		const uploadTask = new LargeFileUploadTask(getClient(), fileObj, uploadSession, options);
+
+		it("Should return a Upload Result object after a completed task with 201 status", async () => {
+			const location = "TEST_URL";
+			const body = {
+				id: "TEST_ID",
+			};
+			const uploadTask = new LargeFileUploadTask(getClient(), fileObj, uploadSession, options);
+			const status201 = {
+				status: 200,
```
zengin, line 170: 201?

zengin: Otherwise this test is exactly the same as the one below.

PR #403, opened 2021-03-15
Defining browser entry, rollup.js, deprecating ImplicitMsalAuthenticationProvider and updating rollup entry

### LINKED ISSUE #394: Use browser field in package.json, remove graph-es-sdk.js file.

### LINKED ISSUE #402: Deprecate ImplicitMSALAuthenticationProvider
[AB#8604](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8604)

nikithauc:
Resolves #394,  resolves #402.

Changes in this PR are mentioned in the #396 design doc.


docs/TokenCredentialAuthenticationProvider.md
```diff
@@ -54,3 +54,19 @@ const client = Client.initWithMiddleware({
...
 ```
+
+The `TokenCredentialAuthenticationProvider` and the `TokenCredentialAuthenticationProviderOptions` are bundled into the `lib/graph-client-tokenCredentialAuthProvider.js` file in an `iife` format.
+
+The browser use of the file is as follows:
+
+```javascript
+// include the script
+<script type="text/javascript" src="<PATH_TO_SCRIPT>/graph-client-tokenCredentialAuthProvider.js"></script>;
+
```
zengin, line 65: suggestion box doesn't allow me to escape three backticks because suggestion itself is using markdown code block.

What I am trying to propose here is that please introduce one code block for html and use html comments for `include the script` line and another code block that for javascript. It would be easier that way for readers of this documentation to copy paste the code into their existing html/javascript.

```suggestion
```html
<!-- include the script -->
<script type="text/javascript" src="<PATH_TO_SCRIPT>/graph-client-tokenCredentialAuthProvider.js"></script>;
```

zengin (approved):

baywet (approved):

PR #408, opened 2021-03-25
Propose getDelta function on GraphRequest

### LINKED ISSUE #407: Add a getDelta function to GraphRequest
# Feature Request

## Is your feature request related to a problem? Please describe

Working with a `/delta` endpoint (like `me/calendarView/delta` in my case) is kind of painful, unless I'm missing something. e.g. here's how I'm extracting the `skiptoken`:

```typescript
const skipToken = new URL(response["@odata.nextLink"]).searchParams.get("$skiptoken") ?? undefined;
```

I have to do the same for `deltatoken`, then check if either is undefined and then proceed accordingly.

## Describe the solution you'd like

See draft PR #408

## Describe alternatives you've considered

If the proposed function "doesn't fit" on `GraphRequest` then I'll just keep it as my own private helper in my code, no worries.

Another approach is to also take a dependency on a vanilla HTTP Client package, set the auth header and call the nextLink/deltaLink directly, but that feels like a bummer.

## Additional context

Taking a step back, the point of this whole SDK is to provide an abstraction over simple HTTP requests to the Graph. You can use it instead of a typical HTTP Client, specifying the auth token, other headers, and URL structure through API calls instead of directly on a request object. But then in the case of a `/deltas` request, the response contains raw URLs to call (`@odata.nextLink` and `@odata.deltaLink`). These are unusable in the context of the SDK!  So you're left calling those links using a vanilla HTTP Client package (and manually setting the auth token), or awkwardly extracting the skip/delta token to pass to the SDK again.

A much nicer experience as an SDK consumer is to have the SDK take care of those details and present a nice API to encapsulate the semantics of a `/delta` request.
 
[AB#8733](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8733)

markfields:
## Summary

This aims to encapsulate semantics of the `@odata/nextLink` and `@odata/deltaLink` scheme for delta requests in a new function `getDelta` on the `GraphRequest` class.

## Motivation

Working with a `/delta` endpoint is kind of painful, unless I'm missing something. e.g. here's how one might extract the `skiptoken`:

```typescript
const skipToken = new URL(response["@odata.nextLink"]).searchParams.get("$skiptoken") ?? undefined;
```

I have to do the same for `deltatoken`, then check if either is undefined and then proceed accordingly.

Or ditch the SDK altogether at this point and craft a plain HTTP request to the next/delta link, digging up the auth token to put on there. Sounds like a bummer.

## Test plan

_TBD - Draft PR for illustrative purposes only.  Will revisit if consensus is reached on the direction of the change._

_I haven't even cloned the repo yet - I just pasted the helper I wrote from my private project in the GitHub web editor and made some tweaks by hand, it does not even build_ :)

## Closing issues

Fixes #407

## Types of changes

## Checklist

_This is a Draft PR, if there's consensus on the change I will follow-up with a proper PR, checking all these boxes_

### DIFF
--- a/src/GraphRequest.ts
+++ b/src/GraphRequest.ts
@@ -53,15 +53,17 @@ export interface URLComponents {
 }
 
 export type DeltaResponse<T> = {
-    tokenType: "skip",
+    linkType: "next",
     skipToken: string,
-    value: T[],
-    nextRequest: () => Promise<DeltaResponse<T>>,
+    followNextLink: () => Promise<DeltaResponse<T>>,
+    body: any,
 } | {
-    tokenType: "delta",
+    linkType: "delta",
     deltaToken: string,
-    nextRequest: () => Promise<DeltaResponse<T>>,
+    followDeltaLink: () => Promise<DeltaResponse<T>>,
+    body: any,
 };
+
 /**
  * @class
  * A Class representing GraphRequest
--- a/src/GraphRequest.ts
+++ b/src/GraphRequest.ts
@@ -52,6 +52,16 @@ export interface URLComponents {
 	otherURLQueryOptions?: string[];
 }
 
+export type DeltaResponse<T> = {
+    tokenType: "skip",
+    skipToken: string,
+    value: T[],
+    nextRequest: () => Promise<DeltaResponse<T>>,
+} | {
+    tokenType: "delta",
+    deltaToken: string,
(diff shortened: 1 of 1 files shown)


nikithauc:
@markfields Thank you for this PR! 

I am closing this PR for now and we can reopen this as discussions and planning on the related issue progress .

PR #429, opened 2021-04-19
Sebastienlevert/fix chaos handler random

### LINKED ISSUE #427: ChaosHandler cannot be used in interactive scenarios
# Bug Report

## Prerequisites

-   [x] Can you reproduce the problem?
-   [x] Are you running the latest version?
-   [x] Are you reporting to the correct repository?
-   [x] Did you perform a cursory search?

## Description

When using ChaosHandler, it should work in interactive scenarios (not only Unit Tests). Right now, if it fails once, it will always fail, which creates a wrong experience in testing throttling handling.

## Steps to Reproduce

1. Inject a ChaosHandler in a list of middleware
2. If the first requests gets throttled, then it will always fail. Making testing an interactive scenario impossible

Every new request should get randomized and potentially succeed.

It's always failing if the initial random identifies the request as a 429.

## Additional Context

## Usage Information

SDK Version - [SDK version you are using]

-   [x] Node (Check, if using Node version of SDK)

> Node Version - 10

-   [x] Browser (Check, if using Browser version of SDK)

> Browser Name - Edge

> Version - 89.0
 
[AB#9063](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9063)

sebastienlevert:
## Motivation

Brings the ChaosHandler to support interactive scenarios. Will allow for randomized requests to fail but also retries attemps to both fail and succeed based on the percentage provided.

## Closing issues

Fixes #427 

## Types of changes

-   [] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/middleware/ChaosHandler.ts
+++ b/src/middleware/ChaosHandler.ts
@@ -77,7 +77,7 @@ export class ChaosHandler implements Middleware {
 
 		if (chaosHandlerOptions.statusCode === 429) {
 			// throttling case has to have a timeout scenario
-			responseHeader.append("retry-after", "300");
+			responseHeader.append("retry-after", "3");
 		}
 
 		return responseHeader;
@@ -129,7 +129,7 @@ export class ChaosHandler implements Middleware {
 		const responseHeader = this.createResponseHeaders(chaosHandlerOptions, requestID, requestDate.toString());
 		const responseBody = this.createResponseBody(chaosHandlerOptions, requestID, requestDate.toString());
 		const init: any = { url: requestURL, status: chaosHandlerOptions.statusCode, statusText: chaosHandlerOptions.statusMessage, headers: responseHeader };
-		context.response = new Response(responseBody, init);
+		context.response = new Response(typeof responseBody === "string" ? responseBody : JSON.stringify(responseBody), init);
 	}
 
 	/**
@@ -141,10 +141,12 @@ export class ChaosHandler implements Middleware {
 	 */
 	private async sendRequest(chaosHandlerOptions: ChaosHandlerOptions, context: Context): Promise<void> {
 		this.setStatusCode(chaosHandlerOptions, context.request as string, context.options.method as RequestMethod);
-		if (!chaosHandlerOptions.statusCode) {
-			await this.nextMiddleware.execute(context);
-		} else {
+		if (Math.floor(Math.random() * 100) < chaosHandlerOptions.chaosPercentage) {
 			this.createResponse(chaosHandlerOptions, context);
+		} else {
+			if (this.nextMiddleware) {
+				await this.nextMiddleware.execute(context);
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/middleware/ChaosHandler.ts
```diff
@@ -141,10 +141,12 @@ export class ChaosHandler implements Middleware {
 	 */
 	private async sendRequest(chaosHandlerOptions: ChaosHandlerOptions, context: Context): Promise<void> {
 		this.setStatusCode(chaosHandlerOptions, context.request as string, context.options.method as RequestMethod);
-		if (!chaosHandlerOptions.statusCode) {
-			await this.nextMiddleware.execute(context);
-		} else {
+		if (Math.floor(Math.random() * 100) < chaosHandlerOptions.chaosPercentage) {
```
nikithauc, line 144: This condition will always be checked even when the user selects 'ChaosStrategy.MANUAL'.  and  as ChaosPercentage defaults to 10 -> 
`this.chaosPercentage = chaosPercentage !== undefined ? chaosPercentage : 10;`

ChaosPercentage indicates the percentage of randomness/chaos in the handle. 

Setting the ChaosPercentage when using the manual strategy might be a confusing convention. 

I feel we should check for the chaos strategy as well and determine the condition.

nikithauc:
@sebastienlevert  Thank you for your contribution!

test/common/middleware/ChaosHandler.ts
```diff
@@ -276,5 +276,17 @@ describe("ChaosHandler.ts", () => {
 			};
 			assert.isDefined(tempChaosHandlerManual["execute"](cxt));
 		});
+
+		it("Should return response for Manual Request Level case 100%", async () => {
```
baywet, line 280: no await in the body, remove the async

baywet (changes requested):
thanks for the contribution @sebastienlevert ! a few minor suggestions

test/common/middleware/ChaosHandler.ts
sebastienlevert: They are all using async without await... Want me to clear all of them in this file?

src/middleware/ChaosHandler.ts
sebastienlevert: Makes sense. I'll do both!

test/common/middleware/ChaosHandler.ts
baywet: sure. thanks to taking this on.

baywet (approved):

src/middleware/ChaosHandler.ts
```diff
@@ -141,10 +141,10 @@ export class ChaosHandler implements Middleware {
 	 */
 	private async sendRequest(chaosHandlerOptions: ChaosHandlerOptions, context: Context): Promise<void> {
 		this.setStatusCode(chaosHandlerOptions, context.request as string, context.options.method as RequestMethod);
-		if (!chaosHandlerOptions.statusCode) {
-			await this.nextMiddleware.execute(context);
-		} else {
+		if (chaosHandlerOptions.chaosStrategy === ChaosStrategy.MANUAL || Math.floor(Math.random() * 100) < chaosHandlerOptions.chaosPercentage) {
```
nikithauc, line 144:
```suggestion
		if ((chaosHandlerOptions.chaosStrategy === ChaosStrategy.MANUAL && !this.nextMiddleware) || Math.floor(Math.random() * 100) < chaosHandlerOptions.chaosPercentage) {
```

PR #437, opened 2021-04-22
Setting the version to 3.0.0-Preview.1

nikithauc:


baywet (commented):
package-lock hasn't been updated as well?

nikithauc:
=

> package-lock hasn't been updated as well?

Thanks for pointing out! Updated the version in two more files.

zengin (approved):

PR #446, opened 2021-05-03
Correct broken links and add reference to auth provider doc

nikithauc:


docs/CustomAuthenticationProvider.md
```diff
@@ -1,6 +1,7 @@
 # Using Custom Authentication Provider
 
-Using preferred choice of Authentication library for authenticating with Microsoft is possible.
+Using preferred choice of Authentication library for authenticating with Microsoft is possible. 
+Learn how to [choose a Microsoft Graph authentication provider based on scenario](https://docs.microsoft.com/en-us/graph/sdks/choose-authentication-providers?tabs=CS) here.
```
MIchaelMainer, line 4: This link doesn't have information about creating a custom authentication provider. I think we should recast this document to be about authentication options, which has the link above. Perhaps rename this topic to authentication.md, have one section about the built TokenCredential, and a second section for the custom option.

nikithauc: The [client creation document](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/docs/CreatingClientInstance.md) specifies how to use the AuthenticationProviders  built in the library. 

I feel that the content of CustomAuthenticationProvider.md  justifies its tittle because we are providing the developer how to implement the `AuthenticationProvider` interface and use a custom provider which is `MyAuthenticationProvider`
 in the example.

I think the tittle of `choose a Microsoft Graph authentication provider based on scenario` can be confusing. 

The link provides more how to choose an authentication flow based on the type of application.
May be we can rename this title to -> "how to authenticate your application" or something else. 

We should also enlist the multiple libraries language wise that are available. Example `msal`, `azure/identity`.

@MIchaelMainer What do think?

PR #453, opened 2021-05-12
Adding `CustomHosts` options

nikithauc:
Related to - https://github.com/microsoftgraph/msgraph-sdk-design/pull/45/files 

Certain workloads error out when an unexpected header is present in the request.

Example - 

In case of `LargeFileUploadTask` in JS library, the upload is as follows:

```
client.api("UPLOAD_URL").put(chunk_to_be_uploaded)
```
The middleware appends request headers which are meant for the Graph API.  

- When I tried uploading to `Outlook`, an error  was caused because of `Authorization` header.
- [LargeFileUploadTask upload to OneDrive caused cors issue due  `sdkversion` telemetry header](https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/265#issue-556249096)

The solution for this is to add a check if the request URL is a Graph endpoint and not send headers to `UPLOAD_URL`.

Adding the headers only for Graph endpoints ignores the case where the developer might have their own test endpoints. 

This PR contains changes to allow developers to provide custom endpoints and update our conditions to modify request headers.


src/GraphRequestUtil.ts
```diff
@@ -65,6 +65,25 @@ export const serializeContent = (content: any): any => {
  * @returns {boolean} - Returns true if the url is a Graph URL
  */
 export const isGraphURL = (url: string): boolean => {
+	return isValidEndpoint(url);
+};
+
+/**
+ * Checks if the url is for one of the custom hosts provided during client initialization
+ * @param {string} url - The url to be verified
+ * @returns {boolean} - Returns true if the url is a for a custom host
```
MIchaelMainer, line 74: * @param {Set<string>} customHostNames - The url to be verified

src/GraphRequest.ts
```diff
@@ -367,12 +367,15 @@ export class GraphRequest {
 		let rawResponse: Response;
 		const middlewareControl = new MiddlewareControl(this._middlewareOptions);
 		this.updateRequestOptions(options);
+		const customHosts = this.config?.customHosts;
```
MIchaelMainer, line 370: I think we should verify that the custom hosts set in the configuration is valid, and if not, throw a client error instructing the developer to provide a host in the expected format. Since they may copy and paste from a URL, I wouldn't be surprised to see a `https://graph.microsoft.com/v1.0` set in the configuration.

baywet (dismissed):

src/GraphRequest.ts
nikithauc: Added a check to throw exception if the hostname is not valid in the config

MIchaelMainer (approved):

PR #460, opened 2021-06-02
Resume stream upload 

### LINKED ISSUE #452: Large File Upload - retry doesn't work for StreamUpload
# Bug Report

## Description

It looks like the retry for the large file upload doesn't work for `StreamUpload`.
When the [`resume`](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/ed06eb73cf8ed819ca2b9938cb077416e3e25334/src/tasks/LargeFileUploadTask.ts#L356) is invoked, in the first step the [`getStatus`](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/ed06eb73cf8ed819ca2b9938cb077416e3e25334/src/tasks/LargeFileUploadTask.ts#L344) method is called in order to update `expiry` and `nextRange` properties - at this step everything is fine. 

After properties are properly updated, the [`upload`](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/ed06eb73cf8ed819ca2b9938cb077416e3e25334/src/tasks/LargeFileUploadTask.ts#L243) is called and then [`sliceFile()`](https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/ed06eb73cf8ed819ca2b9938cb077416e3e25334/src/tasks/LargeFileUploadTask.ts#L252) from the StreamUpload that is causing a problem - it should read the bytes that should be retried, but instead, it's just getting next portion of bytes which seems to be not correct since the bytes that should be retired are not read from the stream.

## Steps to Reproduce

Call `resume` method when the `StreamUpload` is passed to the `LargeFileUploadTask`:
```
function test(): Promise<any> {
   const fileObject = new StreamUpload(readStream, data.fileName, data.fileSize);
   const uploadTask = new LargeFileUploadTask(client, fileObject, uploadSession);

  try {
    await uploadTask.upload();
  } catch(error) {
    await uploadTask.resume();
  }
}
```

  

**Expected behavior:** The upload is properly resumed for the stream data. 

**Actual behavior:** The retry doesn't work for the stream data.

## Usage Information

SDK Version - 3.0.0-Preview.1

 
[AB#9415](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9415)

nikithauc:
fixes #452 

Problem - 
*`StreamUpload.sliceFile` function reads the first `n bytes` from the Stream.
* If an upload fails after the `0 - n` is read from the stream, then on upload resume the `sliceFile` function reads from `n+1` bytes and `0-n` is never uploaded.

Solution -
* Added `ChunkRecord` interface which keeps track of the previous slice and range.
* Compare the new upload range with the previous range and read the bytes accordingly.

### DIFF
--- a/src/tasks/FileUploadTask/FileObjectClasses/StreamUpload.ts
+++ b/src/tasks/FileUploadTask/FileObjectClasses/StreamUpload.ts
@@ -4,12 +4,28 @@ import { GraphClientError } from "../../../GraphClientError";
 import { FileObject, SliceType } from "../../LargeFileUploadTask";
 import { Range } from "../Range";
 
+/**
+ * @interface
+ * Interface to store slice of a stream and range of the chunk.
+ * @property {Buffer} chunk - The slice of the stream
+ * @property {Range} range - The range of the slice
+ */
 interface ChunkRecord {
 	chunk: Buffer;
 	range: Range;
 }
+
+/**
+ * @class
+ * FileObject class for Readable Stream upload
+ */
 export class StreamUpload implements FileObject<Readable> {
+	/**
+	 * @private
+	 * Previous slice information.
+	 */
 	private previousChunk: ChunkRecord;
+
 	public constructor(public content: Readable, public name: string, public size: number) {
 		if (!content || !name || !size) {
 			throw new GraphClientError("Please provide the Readable Stream content, name of the file and size of the file");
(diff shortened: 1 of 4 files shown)


test/node/tasks/StreamUpload.ts
```diff
@@ -51,3 +52,47 @@ describe("StreamUpload", () => {
 		assert.equal(sliceSize, (slice as Buffer).length);
 	});
 });
+
+describe("Stream upload resume", () => {
```
MIchaelMainer, line 56: Do these tests cover scenarios where:
- upload() fails before the bytes are uploaded so resume() uses nextExpectedRange = [0 , rangesize-1] to set nextRange.
- upload() fails after the bytes are uploaded so resume() uses nextExpectedRange = [n , n+rangesize-1] to set nextRange. This assumes we get a response.

src/tasks/FileUploadTask/FileObjectClasses/StreamUpload.ts
```diff
@@ -3,7 +3,29 @@ import { Readable } from "stream";
...
 import { FileObject, SliceType } from "../../LargeFileUploadTask";
 import { Range } from "../Range";
+
+/**
+ * @interface
+ * Interface to store slice of a stream and range of the chunk.
+ * @property {Buffer} chunk - The slice of the stream
+ * @property {Range} range - The range of the slice
+ */
+interface ChunkRecord {
```
MIchaelMainer, line 13: Perhaps call this SliceRecord

src/tasks/FileUploadTask/FileObjectClasses/StreamUpload.ts
```diff
@@ -3,7 +3,29 @@ import { Readable } from "stream";
...
+/**
+ * @class
+ * FileObject class for Readable Stream upload
+ */
 export class StreamUpload implements FileObject<Readable> {
+	/**
+	 * @private
+	 * Previous slice information.
+	 */
+	private previousChunk: ChunkRecord;
```
MIchaelMainer, line 27: This represents a cache of the last attempted upload slice. This should be used when resuming an upload.

test/node/tasks/StreamUpload.ts
MIchaelMainer: I think these cases are the same as resume() always use the uploadUrl from the session to get the nextExpectedRange.

MIchaelMainer (approved):

PR #484, opened 2021-07-19
MsalBrowserAuthenticationProvider 

nikithauc:
Resolves #483.

PR #485, opened 2021-07-22
Removing ImplicitMSAL

nikithauc:
This will have merge conflicts with #484 . To be merged after #484.


package.json
```diff
@@ -1,135 +1,132 @@
...
-		"uglify-es": "^3.3.9"
-	},
-	"peerDependenciesMeta": {
-		"@azure/identity": {
-			"optional": true
-		},
-		"buffer": {
-			"optional": true
-		},
-		"msal": {
```
baywet, line 125: there's an msal entry line 106

src/authentication/msal/MSALAuthenticationProviderOptions.ts
```diff
@@ -10,27 +10,3 @@
  */
 
 import { AuthenticationProviderOptions } from "../../IAuthenticationProviderOptions";
-
```
baywet, line 13: why are we keeping the file at all?

baywet (changes requested):

package.json
nikithauc: Removed this. Thanks for pointing!

src/authentication/msal/MSALAuthenticationProviderOptions.ts
nikithauc: Kept this file with MSALBrowserAuthProviderOptions.  This file will contain options related to MSAL library or libraries (TBD).

baywet (approved):

PR #494, opened 2021-08-06
Fix wrong buffer size calculation.

buptliuhs:
## Summary

The calculation of Buffer size is incorrect (in both v2 and v3).

## Motivation

When using `OneDriveLargeFileUploadTask` to upload small files (usually < 8KB, Buffer.poolSize), buffer size and slice are not calculated correctly, which causes upload failure.

## Example of the issue

When uploading a tiny jpeg file whose size is 747 bytes.
Buffer passed in OneDriveLargeFileUploadTask is:
`{ byteOffset: 5808, byteLength: 747, length: 747, poolSize: 8192 }`

In this case, `byteOffset` is `5808` and `byteLength` is `747`.
The old code calculates size = `byteLength - byteOffset`, which is obviously incorrect.

Because of this, the HTTP header that is used later to upload slices will be like:
```
{
    "content-length": "0",
    "content-range": "bytes 0--5062/-5061",
    ....
}
```
And it causes a 400 response.

It's likely to happen when the buffer size is less than the pollSize. When buffer is large, byteLength is usually 0.

## Test plan

Upload small files (< 8KB) using `OneDriveLargeFileUploadTask`.

## Types of changes

## Checklist

### DIFF
--- a/samples/package-lock.json
+++ b/samples/package-lock.json
@@ -1,1096 +1,1096 @@
 {
-  "name": "samples",
-  "version": "1.0.0",
-  "lockfileVersion": 1,
-  "requires": true,
-  "dependencies": {
-    "@azure/abort-controller": {
-      "version": "1.0.4",
-      "resolved": "https://registry.npmjs.org/@azure/abort-controller/-/abort-controller-1.0.4.tgz",
-      "integrity": "sha512-lNUmDRVGpanCsiUN3NWxFTdwmdFI53xwhkTFfHDGTYk46ca7Ind3nanJc+U6Zj9Tv+9nTCWRBscWEW1DyKOpTw==",
-      "requires": {
-        "tslib": "^2.0.0"
-      }
-    },
-    "@azure/core-asynciterator-polyfill": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@azure/core-asynciterator-polyfill/-/core-asynciterator-polyfill-1.0.0.tgz",
-      "integrity": "sha512-kmv8CGrPfN9SwMwrkiBK9VTQYxdFQEGe0BmQk+M8io56P9KNzpAxcWE/1fxJj7uouwN4kXF0BHW8DNlgx+wtCg=="
-    },
-    "@azure/core-auth": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/@azure/core-auth/-/core-auth-1.3.0.tgz",
-      "integrity": "sha512-kSDSZBL6c0CYdhb+7KuutnKGf2geeT+bCJAgccB0DD7wmNJSsQPcF7TcuoZX83B7VK4tLz/u+8sOO/CnCsYp8A==",
-      "requires": {
-        "@azure/abort-controller": "^1.0.0",
-        "tslib": "^2.0.0"
-      }
-    },
-    "@azure/core-http": {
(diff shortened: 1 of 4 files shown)


nikithauc:
@buptliuhs Thank you for the contribution!

nikithauc:
@buptliuhs 

1. Can you please provide the error and the steps to reproduce the error? 
2. Would it possible to write a unit test identifying this change?

buptliuhs:
hi @nikithauc,

What I did to reproduce the issue is that uploading a tiny jpeg file whose size is 747 bytes.
Buffer passed in OneDriveLargeFileUploadTask is:
{ byteOffset: 5808, byteLength: 747, length: 747, poolSize: 8192 }

In this case, byteOffset is 5808 and byteLength is 747.
The old code calculate size = byteLength - byteOffset, which is obviously incorrect.

Because of this, the http header will be like:
{
    "content-length": "0",
    "content-range": "bytes 0--5062/-5061",
    ....
}

It's likely to happen when the buffer size is less than the pollSize. When buffer is large, byteLength is usually 0.

I'll see how to add a unit test to cover this change.

Cheers,
Tony

buptliuhs:
@nikithauc I have to move some code out of the `create` method to make the change testable. Hope it is okay.

buptliuhs:
hi @nikithauc , are you happy with the change?

nikithauc (dismissed):

nikithauc:
@buptliuhs the changes look good. I changed the base to dev for this PR. 

Can you please resolve this? 

I should have noticed the base setting before. I apologize for the inconvenience.

nikithauc (dismissed):

buptliuhs:
Hi @nikithauc, I was initially looking for a fix in v2, which branch should the change be based on? I can creat another PR to propagate the fix to dev too.

nikithauc:
@buptliuhs Can you please branch off dev? The fix will be reflected in 3.x versions.

buptliuhs:
Hi @nikithauc , I've done the rebase on the branch and now the branch is based on `dev`. Please review.
There are following auto-changes as well, hope they are okay:
- package-lock.json: auto-updated when running `npm install`
- samples/*: auto-updated when running `git commit`

Again, I'm also looking for the same fix on v2 since v3 is still in the preview stage. Is v2 still in maintenance mode? If yes, which branch should I target? Thanks!

Cheers,
Tony

nikithauc:
@buptliuhs 3.0.0 is released - https://www.npmjs.com/package/@microsoft/microsoft-graph-client/v/3.0.0. 2.x version will no longer be seeing any changes.

.gitignore
```diff
@@ -1,6 +1,7 @@
 /.vscode/*
 !/.vscode/launch.json
 !/.vscode/tasks.json
+.idea/
```
nikithauc, line 4: What is the .idea/?

buptliuhs: Oh, I was using Webstorm, .idea/ is similar to .vscode/ in Visual Code. Will remove this.

buptliuhs:
Just did a final tidy-up.

nikithauc (approved):

nikithauc:
@buptliuhs Thank you for this contribution! Merging this right now :)

nikithauc:
@buptliuhs https://www.npmjs.com/package/@microsoft/microsoft-graph-client/v/3.0.1

PR #502, opened 2021-08-14
Set Content-Type explicitly in LargeFileUploadTask.

buptliuhs:
## Summary

In `LargefileUploadTask`, set `Content-Type` to `application/octet-stream` explicitly, otherwise `application/json` will be used by default, which is not correct.

## Motivation

Although using the default `application/json` does not impact the functionality, if a client using the SDK has something like `http-interceptor` to examine the request payload, incorrect `Content-Type` can be misleading, for example, `application/json` suggests `plain/text` in `json` format.

## Test plan

N/A

## Types of changes

## Checklist

### DIFF
--- a/src/tasks/LargeFileUploadTask.ts
+++ b/src/tasks/LargeFileUploadTask.ts
@@ -127,6 +127,7 @@ export class LargeFileUploadTask<T> {
 	 * @async
 	 * Makes request to the server to create an upload session
 	 * @param {Client} client - The GraphClient instance
+	 * @param {string} requestUrl - The URL to create the upload session
 	 * @param {any} payload - The payload that needs to be sent
 	 * @param {KeyValuePairObjectStringNumber} headers - The headers that needs to be sent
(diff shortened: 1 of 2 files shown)


nikithauc:
@buptliuhs  Thanks again for your contribution!! 

I will review this next week.

Question for you, are you facing an error with the current code that is setting the default header to `application/json` ?

buptliuhs:
@nikithauc , No, files can be successfully uploaded even the header is set to `application/json` by default, so I guess the API server just doesn't check it. This is NOT actually impacting the file upload functionality.
However, as I mentioned, it is not uncommon, on the client-side, there is logic like `http-interception` doing things logs, audits, and etc. Having `application/json` in the header is misleading and can cause unexpected behaviours.
No unit test added for now since the existing part of the code is not covered. I can add some (of course need to do some refactoring to make it testable) if it is needed.
Thanks!

nikithauc (approved):

PR #558, opened 2022-01-14
Adding initial thoughts on Kiota JS for Graph

sebastienlevert:
Design specs for the kiota generated version of the JS SDK.


sebastienlevert:
Before we go any further, I wanted to know if this is helpful and hear from both of you @baywet and @nikithauc. It's absolutely a WIP but I'd love to get early feedback! That will help with defining what is required to make technical decisions! Thanks!

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+
+```bash
+npm install @microsoft/msgraph-sdk-typescript --save
+npm install @microsoft/kiota-authentication-azure --save
+```
+
+```typescript
+// App.ts
+
+import { Client, User, Message, BodyType } from "@microsoft/msgraph-sdk-javascript";
```
nikithauc, line 25:
```suggestion
import { Client, User, Message, BodyType } from "@microsoft/msgraph-sdk-typescript";
```

nikithauc: matching this to npm install @microsoft/msgraph-sdk-typescript --save

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+import { DeviceCodeCredential } from "@azure/identity";
+
+const deviceCodeCredentials = new DeviceCodeCredential({
+	tenantId: "b61f9af1-d6cf-4cc0-a6f6-befb38bc00ed",
+	clientId: "bde251a6-0ef9-42a8-a40b-9ad9bb594b2c",
+});
+
+const scopes = ["User.Read", "Mail.Send"];
+
+const graphClient = Client.init({
```
nikithauc, line 36: The user should also be given a choice to  pass serialization implementations

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+const me = await getMe();
+const meRaw = await getMeRaw();
+await sendMail();
+await sendMailRaw();
+
+async function getMe(): Promise<User | undefined> {
+	return await graphClient.me.get();
+}
+
+async function getMe(): Promise<User | undefined> {
```
zengin, line 49: getMeRaw?

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+const scopes = ["User.Read", "Mail.Send"];
+
+const graphClient = Client.init({
+	accessTokenProvider: new AzureIdentityAccessTokenProvider(deviceCodeCredentials, scopes),
+});
+
+const me = await getMe();
+const meRaw = await getMeRaw();
+await sendMail();
+await sendMailRaw();
```
zengin, line 43: could you add a couple of sentences explaining the difference between raw vs regular?

design/kiota-e2e.md
sebastienlevert: I'm really thinking this should be names javascript. The module you are looping in is a Javascript module so I'd stick to this. I'll make is match the npm i.

design/kiota-e2e.md
sebastienlevert: Yes, very fair point. I'll add some pointers.

design/kiota-e2e.md
sebastienlevert: Yes, absolutely. I was trying to go the "main" path here. I don't see a lot of developers providing their own serialization implementations, to be honest. This should be part of the Client Options you can pass it in, though. Same for fetch.

Do you feel we should be going in depth in this scenario of the current `init` is enough for the purpose of the e2e?

design/kiota-e2e.md
sebastienlevert: Absolutely!

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+While working with the Kiota + Javascript teams, it became clear that we needed more context and provide a better vision on how we should be using a kiota-generated SDK. This design document aims at providing clarity and a better understanding on how developers will be using our SDKs and the underlying kiota packages.
+
+## Constraints
+
+Before we jump into the end-to-end walk-through, it's important to set some constraints.
+
+| Type      | Description                                                                                                                                   |
+| --------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| Platforms | Node : Current and Previous LTS (14 / 16)<br /> Web : Edge, Chrome, Firefox and Safari (Latest released version + immediate previous version) |
+| Modules   | Node : CommonJS<br /> Web : ES Module (ES6)                                                                                                   |
```
baywet, line 12: just curious to what data point lead to commonjs for node?

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+
+const deviceCodeCredentials = new DeviceCodeCredential({
+	tenantId: "b61f9af1-d6cf-4cc0-a6f6-befb38bc00ed",
+	clientId: "bde251a6-0ef9-42a8-a40b-9ad9bb594b2c",
+});
+
+const scopes = ["User.Read", "Mail.Send"];
+
+const graphClient = Client.init({
+	accessTokenProvider: new AzureIdentityAccessTokenProvider(deviceCodeCredentials, scopes),
```
baywet, line 37:
```suggestion
	authenticationTokenProvider: new AzureIdentityAuthenticationProvider(deviceCodeCredentials, scopes),
```

And the authentication provider would expose the access token provider to the client under the covers for arbitrary requests but this would be transparent for the consumer

design/kiota-e2e.md
baywet:
```suggestion
async function getRawMe(): Promise<User | undefined> {
```

those PMs resolving comments without making the changes 😋

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+async function getMe(): Promise<User | undefined> {
+	return await graphClient.me.get();
+}
+
+async function getMe(): Promise<User | undefined> {
+	return await graphClient.api("/me").get();
+}
+
+async function sendMail(): Promise<void> {
+	const message: Message = {
```
baywet, line 54: this is still pending the constructor/serialization/backing store discussion

design/kiota-e2e.md
```diff
@@ -0,0 +1,147 @@
...
+
+	return await client.api("/me/sendMail").post({
+		message: message,
+	});
+}
+```
+
+## NodeJS e2e using the Core library
+
+```bash
```
baywet, line 96: maybe I'm missing something but it feels like what we're trying to show here is already covered by the previous example?

baywet (dismissed):

design/kiota-e2e.md
sebastienlevert: Commit incoming 😋

design/kiota-e2e.md
sebastienlevert: Added the comment about it, yes.

design/kiota-e2e.md
sebastienlevert: I made the decision based on current "state" of NodeJS... We're still seeing way more CommonJS modules in the wild today and we need to be where the developers are. Right now, ES Modules are still marked as "experimental" in Node 14 LTS but not in NodeJS 16 LTS... For that reason, I don't want to build and having to refactor based on the changes they could make.

https://nodejs.medium.com/node-js-version-14-available-now-8170d384567e

That being said, the conversation is different with NodeJS 16. This is fully supported and is the path forward.

Now my question to you @nikithauc and @baywet : As we're planning to build both (CJS and ESM), could we focus on ESM for now and add CJS in the future if the data tells us that we should? What would be the downside to build on LTS 16 and see how this works?

design/kiota-e2e.md
sebastienlevert: I'm ok with this if we have an easy access to the access token!

design/kiota-e2e.md
sebastienlevert: Not exactly. Here we're highlighting that the types should be separate from the service library to allow using the same types with the core (without the service library). Very similar, but I feel is required to understand the problem.

design/kiota-e2e.md
baywet: I believe that we should be able add CJS support later should we need it. We're talking about a different TS configuration, some orchestration to build multiple times, and variant entries in the package json file, unless I'm missing something else.

Hence my comment, I do agree with the opinion of focusing on ESM for now, and adding CJS later if data/feedback requires us to do so.

design/kiota-e2e.md
baywet: maybe add some comments in the code to call out the differences, or use a diff code block

design/kiota-e2e.md
nikithauc: @sebastienlevert  We had this discussion here: 
https://github.com/microsoft/kiota/issues/272#issuecomment-885480031 

Also refer to the following which highlight the struggle with ESM only modules and suggest supporting both: 
https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1
https://www.the-guild.dev/blog/support-nodejs-esm

If we are focusing on Node support then we should be focusing on CJS and not ESM.

A good here is node-fetch v3 which migrated to ESM and the move got some heavy feedback https://github.com/node-fetch/node-fetch/issues/1263

sebastienlevert: The `node-fetch` discussion is pretty interesting. How challenging is it to release both? Our focus should be NodeJS for now, but we should be able to "basic" operations within the web, still. If we can simply bundle for both, I'm ok with delivering more value for both audiences (which I understand are very strong and the NodeJS community did not make the full switch to ESM, even if it's supported.

design/kiota-e2e.md
nikithauc: We will still be allowing user to pass just the access token like we do today. 
```
const client = Client.init({
	defaultVersion: "v1.0",
	debugLogging: true,
	authProvider: (done) => {
		done(error, "ACCESS_TOKEN");
	},
});
```

design/kiota-e2e.md
nikithauc: Current `init` is in the core. 

We will need similar factory method for the  `GraphServiceClient` initialization.

design/kiota-e2e.md
nikithauc: The Graph core will continue releasing both as we do today. For the service library I would consider having separate CJS or ESM libraries as an option.

nikithauc:
This could be added in this document too

Tasks constructors such as `PageIterator` and `LargeFileUpload` tasks should accept both `GraphServiceClient` and `GraphCoreClient` 

```
// both should work
const pageIterator = PageIterator(GraphServiceClient, options);
or
const pageIterator = PageIterator(GraphCoreClient, options);
```

design/kiota-e2e.md
```diff
@@ -0,0 +1,215 @@
...
+
+-   `@microsoft/msgraph-sdk-javascript` ([npm](https://www.npmjs.com/package/@microsoft/msgraph-sdk-javascript-core)) - The service library for making fluent calls to Microsoft Graph v1.0.
+-   `@microsoft/msgraph-sdk-javascript-beta` ([npm](https://www.npmjs.com/package/@microsoft/msgraph-sdk-javascript-type)) - The service library for making fluent calls to Microsoft Graph beta.
+
+You can use [npm](https://www.npmjs.com) to install the Microsoft Graph Javascript Service Library :
+
+```Shell
+npm install @microsoft/msgraph-sdk-javascript --save
+npm install @microsoft/msgraph-sdk-javascript-beta --save-dev
+npm install @microsoft/kiota-authentication-azure --save
```
waldekmastykarz, line 35: What's the purpose of this special auth package? Is it just an example or it something that we'll expect developers to install and understand?

design/kiota-e2e.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+### Importing the right functionalities from the Graph Javascript Service Library
+
+```typescript
+// App.ts
+
+import { Client, User, Message, BodyType } from "@microsoft/msgraph-sdk-javascript";
+import { AzureIdentityAuthenticationProvider } from "@microsoft/kiota-authentication-azure";
```
waldekmastykarz, line 44: Ideally, we should avoid introducing yet another special way to do auth that works only in a particular case (in this case the MS Graph JS SDK). Instead, we should strive to use libraries like MSAL and @azure/identity that developers familiar with the Microsoft Identity Platform would already know. The fewer knew and specialistic concepts we introduce the lower the learning curve to start developing for Microsoft 365.

design/kiota-e2e.md
```diff
@@ -0,0 +1,215 @@
...
+const meRaw = await getMeRaw();
+await sendMail();
+await sendMailRaw();
+
+// The types returned by the fluent API should be the same as the .api() area. It should also be the same (or at least very similar) as the current @microsoft/microsoft-graph-types to offer seamless migration.
+async function getMe(): Promise<User | undefined> {
+	return await graphClient.me.get();
+}
+
+// Allowing raw calls (using the .api() method instead of the full fluent API) is important for migration purposes and cases we don't know the resource beforehands (thinking Graph Explorer, mgt-get, etc.)
```
waldekmastykarz, line 82: I understand the migration scenario, but could you please elaborate the case of not knowing the resource upfront?

design/kiota-e2e.md
```diff
@@ -0,0 +1,215 @@
...
+		message: message,
+	});
+}
+```
+
+## NodeJS e2e using the Core library
+
+### Concept
+
+This example brings very similar concepts, but utilizing only the core library. That means that as a developer, we only care about the core capabilities (authentication, serialization, etc.) but we don't need / want the fluent API. This also highlights that types are still important in this case (similarly to how they were important with the v3 of the SDK using our Typings).
```
waldekmastykarz, line 134: Since core functionality is also available in the service library, why do we want to expose the core library as a separate package? The problem I see with that is that it offers developers another choice that we'll need to explain so that developers can understand when to use which library and what are the trade-offs and limitations of each package. It could also lead to confusion if developers install the core library but want to try to use the fluent API and are stuck because it won't work, not knowing that for that they need a different package. So rather than offering two separate packages, I'd suggest that we offer just the service library with the fluent API being the recommended way of calling Graph. If need be, we offer developers to use `.api()` as an escape hatch, but both options are already available in the same package. As developers gradually upgrade their app from the raw to fluent API to benefit from the design-time typesafety features, they can do so without any additional dependencies.

design/kiota-e2e.md
baywet: this would be the authentication provider to use when you want to use Azure Identity. So most likely for SPAs, or nodeJS apps.
For SPFx/TeamsFx and other "platforms" this package would be different.

design/kiota-e2e.md
baywet: this is just a wrapper around Azure identity, not a new auth lib

design/kiota-e2e.md
baywet: what do you mean by not knowing the resource upfront?

design/kiota-e2e.md
baywet: service libraries will be a layer on top of the core lib, which will offer the api() method like it does today. So effectively the consumers only will need to install the service lib they want to use, plus the side-car for authentication depending on their "platform" and they'll see everything, the api method and the fluent style APIs.

design/kiota-e2e.md
waldekmastykarz: Since MSAL already solved auth for SPAs and Node.js apps, couldn't we reuse what's there already?

design/kiota-e2e.md
waldekmastykarz: Right, but developers don't know that and it requires us to create and maintain a separate set of samples, guidance, tutorials, etc and them to learn one way of doing auth with MSAL and another with our SDK. Is there a way for us to use what's already in MSAL so that we can have a consistent auth story across all Microsoft cloud?

design/kiota-e2e.md
waldekmastykarz: Oh, so you mean cases where the request is built on runtime rather than design-time, correct?

design/kiota-e2e.md
waldekmastykarz: Understood, so theoretically we could choose to expose a single package (service library that does both fluent and raw API) and keep the core library internal and not something we'd ever mention to developers, right?

design/kiota-e2e.md
baywet: The decision to align on Azure Identity to provide consistency across Microsoft clouds (azure, office 365, etc...) was taken prior to Kiota work, almost 2 years ago now. In most cases, azure identity uses MSAL under the covers and adds value on top of it. The JavaScript SDK is the only "special" SDK in the sense that it might run in platforms that include some authentication mechanism (SPFx/TeamsFX). The current version of the SDK already supports Azure Identity in facts, https://github.com/microsoftgraph/msgraph-sdk-javascript/tree/dev/src/authentication/azureTokenCredentials

baywet: And to expand on that, the reason why we didn't strip MSAL all together was because at the time azure identity was limited to node scenarios, which I don't believe is the case anymore.

sebastienlevert: I agree, naming should be better here. If we were more specific with something like `@microsoft/kiota-provider-azure`. We would have the same for `@microsoft/kiota-provider-sharepoint`, `@microsoft/kiota-provider-teamsfx`, `@microsoft/kiota-provider-electron` allowing developer to focus on their business logic, doing the boilerplate code for them and them passing in the right context to allow seemless authentication (a TokenCredential for Azure Identity, a UserCredential for TeamsFx, a context for SharePoint, etc.)

design/kiota-e2e.md
sebastienlevert: Yes, in this example, the response type would be any/unknown. But I believe our developers will want to have a typings experience even using raw http calls. Similar to how you can leverage the SDK today.

design/kiota-e2e.md
baywet: which doesn't solve for SPFx/TeamsFx/Electron and for other authentication mechanisms or libraries. The role of azure identity/msal is to unify the authentication experience for any API that lives behind Microsoft Identity Platform in scenarios where you don't already have a token. The role of those kiota wrappers is to abstract away authentication so the client is not tied to any specific implementation/scenario/mechanism. Two different concerns that go hand in hand.

design/kiota-e2e.md
sebastienlevert: I don't know if we would keep it internal, but definitely not publicize its use. We believe having the core is critical to some advanced scenarios. When using the service library (which will be the preferred and documented approach), you would not need to add the core library explicitely.

baywet: correct, to the exception of the developers who don't want the fluent stuff AT ALL, and just want the core lib. This population would install only core. But that's a niche segment, we'll have documentation for that, but we won't be advertising for it.

design/kiota-e2e.md
nikithauc: @sebastienlevert  do we have a confirmation that azure/identity works well for browser and we do not need an msal auth provider?

Also something to take a note of is that azure/identity does not provide a CDN source.

design/kiota-e2e.md
nikithauc: @sebastienlevert  I suggest we should decide on the publicity of the core and service library based on the usage of the audience and this can be different for different languages. JS being a scripting language is used by a  wide range of users including non-developers. It would be cool to present the libraries as options that the user can pick based on their usage and comfort level.

design/kiota-e2e.md
waldekmastykarz: > The role of those kiota wrappers is to abstract away authentication so the client is not tied to any specific implementation/scenario/mechanism.

And that's exactly the problem because we're introducing yet another abstraction layer. So on top of identity APIs, we have MSAL, @azure/identity and now Kiota providers which are in total 4 different ways to do auth that developers need to know about and understand the benefits and trade-offs of each one of them.

I wonder if the concern about using different providers with SPFx, Teamsfx and other toolchains is relevant here, because these toolchains expose an authenticated version of the Graph client. Yes, it's important that the Graph SDK supports being instrumented with information coming from these toolchains, but developers using the Graph SDK will never see that. They will only see the special way of authenticating the SDK using the Kiota provider that applies only to the Graph SDK. If they use some other Azure SDK or raw REST calls, they will need to use MSAL/@azure identity there, which is a poor experience. What if support for authenticating toolchains was something that we handle internally but externally, for developers using the Graph SDK, they'd be able to use MSAL/@azure identity just as they do in any other Azure SDK or raw REST calls?

design/kiota-e2e.md
waldekmastykarz: @sebastienlevert if we're to provide providers for the different toolchains, what will that mean for the version matrix between the SDK and the different toolchains? What if we at Microsoft, or the community builds another toolchain: what will be the process to onboard it to Kiota to offer a consistent experience?

design/kiota-e2e.md
waldekmastykarz: If you use a raw HTTP call but know the URL upfront, then by extension you'll know the shape of data returned by the API and can cast the response to the correct type, correct? I think the challenge is the odd case where you construct the URL on runtime and you don't know what data you'd be getting, but I don't think that's a common usecase.

design/kiota-e2e.md
waldekmastykarz: > It would be cool to present the libraries as options that the user can pick based on their usage and comfort level.

If you look across all of Microsoft 365, we already ask people to choose between many UI libraries, SDKs, types of apps, toolchains and tools. Adding yet another choice offers additional flexibility, but also requires us to properly explain its ramifications, constraints, recommendations, samples, tutorials and guidance. It also increases the risk of people choosing the wrong thing and having to back track if they find it's not what they need. We should consider this when deciding how to expose the SDKs.

sebastienlevert: I absolutely 100% agree with @waldekmastykarz here. Our core is a tool we use to avoid duplication of code and allow common features to be shared with v1 and beta. I would argue that we should not publicize the core libraries (except for very advanced scenarios) as they are getting the exact same value via the service library. Our tree-shaking should deliver what the user needs, and this should be delivered via our service library.

design/kiota-e2e.md
sebastienlevert: This would not land in kiota. This could be a "community-led" kiota-supported authentication provider that anybody could leverage. It would not live in kiota (would leverage its set of interfaces as a dependency), but in a separate repo. For Microsoft, it could live in kiota if we believe there is a need / opportunity for us to deliver it. Right now, we are thinking to bring some of the MGT-specific providers (SPFx, TeamsFx, Electron) natively to kiota, so any kiota-built SDK would support these platforms also.

design/kiota-e2e.md
sebastienlevert: I understand your point. Do you think we should literally avoid the types here? It feels there is value for developers with a preference with using endpoint vs. the fluent API...

design/kiota-e2e.md
waldekmastykarz: > I didn't know you could mention an org on GitHub, I wonder whether it notifies anybody when you do that...

That wasn't my intention. It was meant to denote the `@azure/identity` package 😊

design/kiota-e2e.md
waldekmastykarz: Perhaps the consistency is not that relevant after all. Similarly how you can work with eslint plugins: some are provided by folks behind eslint, others come from the community. As long as there is a common contract and documentation how to build and set them up, their location and name might not matter.

> Right now, we are thinking to bring some of the MGT-specific providers (SPFx, TeamsFx, Electron) natively to kiota, so any kiota-built SDK would support these platforms also.

Do you have any example how this would be beneficial? Right now our toolchains are tied to a specific platform, eg. you can use SPFx, MGT or Teamsfx just with JavaScript, so what would be the benefit of having providers for these toolchains in Kiota if in the end they can work with just one provider and one Kiota SDK?

design/kiota-e2e.md
waldekmastykarz: Do you mean: "what if folks wanted to achieve what they're having today on v3, with v4 but without loading the fluent API, and having just the core functionality and typings"?

design/kiota-e2e.md
sebastienlevert: Goal would be to use these providers with "any" Kiota-generated SDK. Any SDK could benefit from SPFx integration, same with TeamsFx for instance.

design/kiota-e2e.md
sebastienlevert: Exactly!

waldekmastykarz: If we want to support this scenario, wouldn't that mean that we need to offer two SDKs (core and service) after all which brings us back to the original argument?

sebastienlevert: It's not something we'd publicize that much, but I would say yes. This needs to be supported for these advanced scenarios. I'm happy to park this idea and listen for feedback though... Which would allow us to focus on a single path (the service library) assuming it delivers all the underlying core features (which it should).

Issue #529, opened 2021-11-10
Error: The value of parameter 'StartDateTime'/'EndDateTime' is invalid.


moinkhanif:
# Bug Report

## Prerequisites

-   [x] Can you reproduce the problem?
-   [x] Are you running the latest version?
-   [x] Are you reporting to the correct repository?
-   [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

## Description

[Description of the bug or feature]

I was following the [Microsoft Graph steps](https://docs.microsoft.com/en-us/graph/tutorials/node?tutorial-step=4) and stumbled upon this issue at step 5 of "Get calendar events from Outlook". 
The page `/calender` is supposed to return a JSON value. However, I get the following error:
```json
[...]

PR #625, opened 2022-02-10
Update GraphRequest.ts

acristu:
quickfix for https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/529

### DIFF
--- a/src/GraphRequest.ts
+++ b/src/GraphRequest.ts
@@ -224,14 +224,14 @@ export class GraphRequest {
 		if (Object.keys(urlComponents.oDataQueryParams).length !== 0) {
 			for (const property in urlComponents.oDataQueryParams) {
 				if (Object.prototype.hasOwnProperty.call(urlComponents.oDataQueryParams, property)) {
-					query.push(property + "=" + urlComponents.oDataQueryParams[property]);
+					query.push(property + "=" + encodeURIComponent(urlComponents.oDataQueryParams[property]));
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


sebastienlevert:
Thanks for your contribution @acristu! We will be reviewing and merge when ready! This will be available in our next version of the SDK!

acristu:
no pb, sry this was in a hurry, don't have time to run all tests and make a proper PR

actually on my end I just modified the js file dirrectly in node_modules, I need to do a quick PoC ... so for me this workaround ia sufficient for now...

nikithauc:
@acristu Thank you so much for this contribution! 

Unfortunately at present, the library expects an encoded URI to be passed as a parameter. We have customers passing in an encoded URI and I would recommend this as a workaround now. 

I labelled this PR as a breaking change considering cases of encoding already encoded URLs which would break some customers.

acristu:
Hi @nikithauc , thanks for the reply, so this means the 3min Graph intro node.js app should be fixed to send startDateTime/endDateTime in URI encoded format ?

nikithauc:
> Hi @nikithauc , thanks for the reply, so this means the 3min Graph intro node.js app should be fixed to send startDateTime/endDateTime in URI encoded format ? I was following the intro when I hit this issue ... the downloaded app did not work for me ...

Thanks for bringing this to attention! Let me verify this and make the necessary updates:)

nikithauc:
#823

PR #1026, opened 2022-10-27
Build pipeline 

nikithauc:


azure-pipelines.yml
```diff
@@ -0,0 +1,33 @@
...
+# https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
+
+
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '16.x'
```
MIchaelMainer, line 13: Is this the only version spec to use?

MIchaelMainer (changes requested):
Does build or test run linter?
Also, can you add CredScan and Component Governance tasks as the first two tasks?

Please name this pipeline with the `name: ` field. Please change the file name to be more descriptive of what this pipeline does. Then you can use `$(BuildDefinitionName)` as part of the `name: ` so you match the filename with the name that shows up in the Azure Pipelines UI.

Please be explicit about whether this is supposed to run on PR and/or commits. Give path specs that limit the type of changes that should result in a new build. Give branch specs as well since we don't want this to run on all branches. Here's some [reference](https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/blob/master/pipelines/validatePR.yml).

azure-pipelines.yml
```diff
@@ -0,0 +1,33 @@
...
+  displayName: 'Install Node.js'
+
+- script: |
+    npm install
```
baywet, line 17:
```suggestion
    npm ci
```

https://dev-tips.com/nodejs/difference-between-npm-install-and-npm-ci

azure-pipelines.yml
baywet: A matrix could be interesting here https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job-strategy?view=azure-pipelines

azure-pipelines.yml
```diff
@@ -0,0 +1,33 @@
...
+    npm run build
+    npm run test
```
baywet, line 19: if your setup allows for a report to be exported, it might be worth importing the result in ADO to get more visibility over the test results and duration.
https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=trx%2Cyaml

baywet (changes requested):

azure-pipelines.yml
gavinbarron: TIL about npm ci TY @baywet

azure-pipelines.yml
gavinbarron: +1 to this, being able to see what tests are failing on a run without needing to trawl the logs is really valuable.

.azure-pipelines/ci-build-production.yml
```diff
@@ -0,0 +1,90 @@
...
+
+pr: none
+
+
+stages:
+- stage: "Test"
+  jobs:
+  - job:
+    pool:
+      vmImage: windows-latest
```
baywet, line 21: I think that being on windows was a requirement for test reports a while back but I believe it's not the case anymore. Maybe worth trying again with ubuntu to get the performance benefits.

baywet (dismissed):
Thanks for taking in the suggestions!

.azure-pipelines/ci-build-production.yml
nikithauc: The security tasks use a window image

baywet: ah right, I forgot about that... thanks for pointing it out.

MIchaelMainer (dismissed):

MIchaelMainer (dismissed):

.azure-pipelines/ci-build-production.yml
```diff
@@ -0,0 +1,89 @@
...
+    - task: NodeTool@0
+      inputs:
+        versionSpec: $(NODE_VERSION)
+      displayName: 'Install Node.js'
+
+    - script: npm ci
+    - script: npm run build
+    - script: npm run test
+    - script: npm run karma
+      displayName: 'Test the branch'
```
gavinbarron, line 42: Can we please have `displayName` set for each of these script steps?

gavinbarron (changes requested):

.azure-pipelines/ci-build-production.yml
nikithauc: Done

gavinbarron (approved):
🚀

PR #1131, opened 2023-01-10
Mocha version update

nikithauc:
- Taking the latest version of mocha with is not compatible with `esm` library.
- Added a simple tiny project to test if a ESM project with "type:module" compiles fine with the Graph JS library.


gavinbarron (approved):
LGTM

PR #1133, opened 2023-01-12
Add header property to Graph error object

### LINKED ISSUE #1113: No claim present in 'microsoft-graph-client' package's create api's 401 unauthorized response
# Bug Report

## Prerequisites

-   [ ] Can you reproduce the problem?
- Yes
-   [ ] Are you running the latest version?
- Yes
-   [ ] Are you reporting to the correct repository?
- Yes
-   [ ] Did you perform a cursory search?
- Yes

For more information, see the `CONTRIBUTING` guide.

## Description

[Description of the bug or feature]

When hitting below URL

[https://graph.microsoft.com/v1.0/me/drive/root:/Transcribed%...](https://graph.microsoft.com/v1.0/me/drive/root:/Transcribed%20Files/Recording.mp3:/CreateUploadSession)

 
We are getting 401 Unauthorized error while onedrive upload  but there is no claim present in it, which we can see in browser network traces. Since this claim is missing in error response body, we are not able to request New token.
According to our understanding only the response body is coming in case of exception while we need response headers as well to read claim challenge value in current case

The reason for 401 Unauthorized is fine as Server want Client to reauthenticate itself and fetch Fresh token but issue is error body is missing the claim value.  This is leading to we not being able to refresh Access Token.

As can be seen in below attached network trace, Http response body is having claim value under WWW-Authenticate header but same is not getting passed in error response, requesting you to kindly check the above mentioned and help.

Network Traces  and error response body screenshot attached below 
**Console Errors:** [Is there any console error]

**Screenshots:** [If applicable, add screenshots to help explain your problem]
image
image

## Steps to Reproduce

1) Sign in with a account in win32 word.
2) Try to launch 'Transcribe' feature from the Home tab
3) Click 'Live Recording' and generate a transcript.
4) Change the password of the account.
5) Try to generate transcript again in the same pane session.
6) The onedrive upload will fail with 401 unauthorized error.

We should be able to get the transcript after handling claim challenge in the code.
We got 401 unauthorized error.

## Additional Context

## Usage Information
Request ID - Value of the `requestId` field if you are receiving a Graph API error response

SDK Version - [SDK version you are using]

-   [ ] Node (Check, if using Node version of SDK)

> Node Version - [The version of Node you are using]

-   [ ] Browser (Check, if using Browser version of SDK)

> Browser Name - [The name of Browser that you are using for SDK]

> Version - [The version of the browser you are using]

nikithauc:
fixes #1113

### DIFF
--- a/shims.d.ts
+++ b/shims.d.ts
@@ -19,3 +19,5 @@ interface NodeStream {
     read(size?: number): any;
     on(event: string | symbol, listener: (...args: any[]) => void): this;
 }
+
+interface Headers{}
--- a/src/GraphError.ts
+++ b/src/GraphError.ts
@@ -42,7 +42,7 @@ export class GraphError extends Error {
 	 */
 	public date: Date;
 
-    public headers?: Record<string, string>;
+    public headers?: Headers;
 
 	/**
 	 * @public
--- a/src/GraphError.ts
+++ b/src/GraphError.ts
@@ -42,7 +42,7 @@ export class GraphError extends Error {
 	 */
 	public date: Date;
 
-    public headers?: Headers;
+    public headers?: Record<string, string>;
 
 	/**
 	 * @public
--- a/src/GraphError.ts
+++ b/src/GraphError.ts
@@ -42,6 +42,8 @@ export class GraphError extends Error {
 	 */
 	public date: Date;
 
+    public headers?: Headers;
+
 	/**
 	 * @public
 	 * A member holding original error response by the graph service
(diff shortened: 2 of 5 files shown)


shims.d.ts
```diff
@@ -19,3 +19,5 @@ interface NodeStream {
     read(size?: number): any;
     on(event: string | symbol, listener: (...args: any[]) => void): this;
 }
+
+interface Headers{}
```
gavinbarron, line 23: Why do we need an empty interface defined in here? Should it have some properties/methods defined?

nikithauc: The empty interface is in case the customer doesn't have the lib: "dom" option in their code. Then this will throw an error  on compile. 
Empty interfaces is to make sure that the code compiles and merges with any existing definition.

gavinbarron: Thanks for the context here, much appreciated, and this now makes sense to me.

PR #1368, opened 2023-07-21
feat: use 500 as default status code since it's an unhandled error

teofilomonteiro:
## Summary

If an unhandled error boils up, the statusCode by default is -1.  

## Motivation

Explain the **motivation** for making this change. What existing problem does the pull request solve?

## Test plan

## Closing issues

Fixes #

## Types of changes

## Checklist

-   [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/GraphErrorHandler.ts
+++ b/src/GraphErrorHandler.ts
@@ -94,11 +94,11 @@ export class GraphErrorHandler {
 	 * To get the GraphError object
 	 * Reference - https://docs.microsoft.com/en-us/graph/errors
 	 * @param {any} [error = null] - The error returned by graph service or some native error
-	 * @param {number} [statusCode = -1] - The status code of the response
+	 * @param {number} [statusCode = 500] - The status code of the response. Default is 500 as internal server error
 	 * @param {GraphRequestCallback} [callback] - The graph request callback function
 	 * @returns A promise that resolves to GraphError instance
 	 */
-	public static async getError(error: any = null, statusCode = -1, callback?: GraphRequestCallback, rawResponse?: Response): Promise<GraphError> {
+	public static async getError(error: any = null, statusCode = 500, callback?: GraphRequestCallback, rawResponse?: Response): Promise<GraphError> {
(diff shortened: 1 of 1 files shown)


sebastienlevert:
Thanks for the contribution @teofilomonteiro. In this case, the error you are getting back is probably a client error rather than a server side error. Assigning it to 500 by default wouldn't true in this case as there was an error in sending the request, not in receiving the response. 

Can you expand on the scenario and how it happens in your app? Thanks!

teofilomonteiro:
@sebastienlevert An example I can give you is when your credentials are incorrect. When you make a request the error thrown comes with a status code -1. This is not the correct behaviour. 

Also if you are using a framework like nestjs if you throw an error with status code -1, you break the app. Also sending a -1 is strange. Maybe 500 is better since it was an internal error.

sebastienlevert:
So you are saying that sending an invalid token / without a token results in a -1 in the Graph Response?

sebastienlevert:
If you share a small repro, we'd love to investigate using this. Based on the logic in the Error handler, I don't think we would be missing a status coming from the server for instance.

teofilomonteiro:
Thanks @sebastienlevert! It's a super easy setup.

```typescript
import { ClientSecretCredential } from '@azure/identity';
import { Client } from '@microsoft/microsoft-graph-client';
import { TokenCredentialAuthenticationProvider } from '@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials';

async function main() {
  const tenantId = 'valid_tenant_id';
  const clientId = 'valid_client_id';

  const credential = new ClientSecretCredential(tenantId, clientId, 'wrong_client_secret');
  const authProvider = new TokenCredentialAuthenticationProvider(credential, {
    scopes: ['.default'],
  });

  const client = Client.initWithMiddleware({
    debugLogging: false,
    authProvider,
  });

  await client.api('/domains').get();
}

main();
```

This should replicate the issue 👍  In the end

PR #1403, opened 2023-08-14
chore: remove unsupported version of node (12) from matrix

MIchaelMainer:


baywet (changes requested):
We should at least add 18, maybe even 20. And maybe remove 14

MIchaelMainer:
> We should at least add 18, maybe even 20. And maybe remove 14

14 still has a bit more life in it.

baywet (changes requested):
we should probably align the github workflow with that.
And maybe update the policy?
Also have you tried running that workflow? a couple of weeks ago when I tried node 20 for kiota typescript I was getting failures it was impossible to install

sebastienlevert:
What is the impact on calling developers to up our target? Can we impact existing customers? Is that a breaking change?

If we look at https://nodejs.dev/en/about/releases/, 18 and 20 should be the one supported.

baywet:
IMHO yes

sebastienlevert (changes requested):
Can we make this 16-20?

sebastienlevert:
Though, this could mean making it a breaking change for our users. Should we delay with v4 when the work on Kiota is done?

koros:
There were some failing tests after the upgrade so I have addressed those ones on a new PR: https://github.com/microsoftgraph/msgraph-sdk-javascript/pull/1405

Issue #1513, opened 2023-10-30
Tests don't run


altano:
# Bug Report

## Prerequisites

-   [x] Can you reproduce the problem?
-   [x] Are you running the latest version?
-   [x] Are you reporting to the correct repository?
-   [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

## Description

I'm trying to contribute but I can't get the tests to run.

**Console Errors:** 

`npm run test`:

<img width="491" alt="image" src="https://github.com/microsoftgraph/msgraph-sdk-javascript/assets/1009/c61b7617-a22c-4c9a-af49-fae5d26f573a">

## Steps to Reproduce

1. Clone this repo
2. npm install && npm run build
3. npm run test

**Expected behavior:** Tests run, failures are reported

**Actual behavior:** tests never run

NOTE: I tried `npm run test:cjs` and `npx mocha` with the same results

## Additional Context

I'm creating failures to make sure the tests SHOULD have some output (the failure should be reported, at the least) and nothing:

```ts
describe("RetryHandlerOptions.ts", () => {
	describe("Constructor", () => {
		it("just fails", () => {
			expect.fail("Hi");
		});
```

I'm on nodejs@18.12.0 and npm@8.19.2

## Usage Information

n/a

PR #1514, opened 2023-10-30
increase default retry delay

### LINKED ISSUE #978: SDK doesn't appear to use Retry-After on 429s
# Bug Report

## Prerequisites

-   [X] Can you reproduce the problem?
-   [X] Are you running the latest version?
-   [X] Are you reporting to the correct repository?
-   [X] Did you perform a cursory search?

## Description

During testing of a tool to simulate 429s on the network layer (i.e. outside of the JavaScript library) it was observed that a when a Retry-After or 5s was returned to the client application the SDK issued a retry request before the 5s window had elapsed

**Console Errors:** N/A

**Screenshots:**

## Steps to Reproduce

1. Use the ChaosProxy sample using the sdk and with the proxy running with 99% failure rate and only 429s as the allowed errors
2. see that the a retry is sent before the 5s wait period is finished

SDK should not be retrying until after the number of seconds provided in the `Retry-After` header

SDK retries the request at 3s and eventually succeeds due to the exponential backoff behavior

## Additional Context

Chaos Proxy is here: https://github.com/microsoftgraph/msgraph-chaos-proxy (Microsoft private repo at present)

## Usage Information
Request ID - N/A

SDK Version - script link to https://cdn.jsdelivr.net/npm/@microsoft/microsoft-graph-client/lib/graph-js-sdk.js so I assume latest

-   [x] Browser (Check, if using Browser version of SDK)

> Browser Name - Edge

> Version - 105

altano:
## Summary

Creates new defaults for retrying delays:

| Thing               | Old Value   | New Value      | Reasoning                                                                                                                 |
| ------------------- | ----------- | -------------- | ------------------------------------------------------------------------------------------------------------------------- |
| DEFAULT_MAX_RETRIES | 3           | 12             | 2^12 seconds is just over 1 hour, which should be enough for OneNote                                                      |
| MAX_DELAY           | 180 seconds | 3_600 (1 hour) | This makes it so that retry 12+ only wait 1 hour between retries instead of continuing to backoff the delay exponentially |
| MAX_MAX_RETRIES     | 10          | 64             | This needs to be higher than DEFAULT_MAX_RETRIES but the choice of 64 is arbitrary                                        |

## Motivation

Some of the graph APIs do not provide the `retry-after` header in HTTP 429 throttling responses. The default behavior of this SDK is to retry 3 times, with exponential backoff. The problem with this approach is that the max retry of 3 attempts (with a delay of ~8 seconds) just isn't enough for some APIs. For example, OneNote's API will potentially return 429s for 1 hour.

See [this GitHub issue](https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/978) for more discussion.

## Test plan

I was hoping to write unit tests for this but I can't get the tests to run (https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/1513).

## Closing issues

Fixes #978

## Types of changes

## Checklist

### DIFF
--- a/src/middleware/options/RetryHandlerOptions.ts
+++ b/src/middleware/options/RetryHandlerOptions.ts
@@ -37,21 +37,21 @@ export class RetryHandlerOptions implements MiddlewareOptions {
 	 * @static
 	 * A member holding default maxRetries value
 	 */
-	private static DEFAULT_MAX_RETRIES = 3;
+	private static DEFAULT_MAX_RETRIES = 12;
 
 	/**
 	 * @private
 	 * @static
-	 * A member holding maximum delay value in seconds
+	 * A member holding maximum delay value (1 hour) in seconds
 	 */
-	private static MAX_DELAY = 180;
+	private static MAX_DELAY = 3_600;
 
 	/**
 	 * @private
 	 * @static
 	 * A member holding maximum maxRetries value
 	 */
-	private static MAX_MAX_RETRIES = 10;
+	private static MAX_MAX_RETRIES = 64;
 
 	/**
 	 * @public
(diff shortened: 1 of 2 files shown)


altano:
@microsoft-github-policy-service agree

PR #2000, opened 2026-06-15
fix: prevent token leak via URL userinfo host confusion

MIchaelMainer:
## Summary

 Changes Made

  1. src/GraphRequestUtil.ts — isValidEndpoint (root cause fix)

  Replaced the hand-rolled indexOf(":") host extraction with new URL() parsing:

   - Rejects non-HTTPS URLs
   - Rejects any URL containing userinfo (username/password) — blocks the host-confusion attack
   - Uses parsedUrl.hostname (which correctly resolves the actual host, excluding userinfo and port)

  2. src/GraphRequest.ts — parsePath (defense-in-depth)

  Added new URL() validation at the entry point of path parsing:

   - Rejects malformed URLs with a clear error
   - Throws if userinfo is present — prevents credentialed URLs from ever entering the request pipeline
   - After validation, uses the original string for host/path extraction (preserving encoding and port for backward compatibility)

## Motivation

A vulnerability in the Microsoft Graph SDK for JavaScript allows an attacker to exploit a host-classification bug in the `isValidEndpoint` function, causing the SDK to misclassify attacker-controlled URLs as valid Microsoft Graph endpoints. This results in the SDK attaching and sending the caller's Microsoft Graph bearer token to the attacker's host. The issue is confirmed in version 3.0.7 and earlier, specifically when using the `node-fetch` runtime. The root cause is improper URL parsing that fails to correctly identify the real host when userinfo is included in the URL. The vulnerability can lead to the exposure of sensitive tokens, which can be replayed against Microsoft Graph services within the token's granted scopes.

## Test plan

Regression tests added

   - test/common/core/GraphRequestUtil.ts — tests for isGraphURL and isCustomHost covering all attack vectors
   - test/common/core/urlParsing.ts — tests that parsePath throws on userinfo URLs

### DIFF
--- a/src/GraphRequest.ts
+++ b/src/GraphRequest.ts
@@ -132,15 +132,25 @@ export class GraphRequest {
 	private parsePath = (path: string): void => {
 		// Strips out the base of the url if they passed in
 		if (path.indexOf("https://") !== -1) {
-			path = path.replace("https://", "");
+			// Use structured URL parsing to validate the URL and reject userinfo
+			let parsedUrl: URL;
+			try {
+				parsedUrl = new URL(path);
+			} catch {
+				throw new GraphClientError("Unable to parse the URL: " + path);
+			}
+			// Reject URLs with userinfo to prevent host-confusion attacks
+			if (parsedUrl.username !== "" || parsedUrl.password !== "") {
+				throw new GraphClientError("URL cannot contain user credentials: " + path);
+			}
 
-			// Find where the host ends
-			const endOfHostStrPos = path.indexOf("/");
+			// Extract host from the original string using the first "/" after "https://"
+			const withoutScheme = path.substring("https://".length);
+			const endOfHostStrPos = withoutScheme.indexOf("/");
 			if (endOfHostStrPos !== -1) {
-				// Parse out the host
-				this.urlComponents.host = "https://" + path.substring(0, endOfHostStrPos);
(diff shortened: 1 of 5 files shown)


src/GraphRequest.ts
```diff
@@ -132,15 +132,25 @@ export class GraphRequest {
...
-			path = path.replace("https://", "");
+			// Use structured URL parsing to validate the URL and reject userinfo
+			let parsedUrl: URL;
+			try {
+				parsedUrl = new URL(path);
+			} catch {
+				throw new GraphClientError("Unable to parse the URL: " + path);
+			}
+			// Reject URLs with userinfo to prevent host-confusion attacks
+			if (parsedUrl.username !== "" || parsedUrl.password !== "") {
```
adrian05-ms, line 143: is it possible for username to be undefined? what happens if it does?

MIchaelMainer: Good question. It turns out, no, it can't:

```js
node -e "const u = new URL('https://example.com'); u.username = undefined; console.log('username:',
JSON.stringify(u.username), typeof u.username); u.username = null; console.log('username:',
JSON.stringify(u.username), typeof u.username); console.log('href:', u.href)"
```

Output:
```
username: "undefined" string
username: "null" string
href: https://null@example.com/   
```

gavinbarron (approved):

